PR #6, opened 2021-07-12
Elens

phermes:
add hollow electron lens to xtrack.


giadarol (approved):

giadarol:
@phermes Thanks a lot!!!!

Before merging this, it would be useful to have the e-lens covered by our test suite, to make sure that we don't break anything with future changes.
Would you mind adding a small test to test suite? 
For example you could compare the e-lens implementation in xtrack and xline. You can use [this test](https://github.com/xsuite/xtrack/blob/main/tests/test_elements.py) as an example.

Thanks in advance!

phermes:
Sure, that's a good idea. Will add it to the test.

phermes:
Done !

giadarol (approved):

PR #19, opened 2021-07-30
Tracking with linear transfer matrices

xbuffat:


giadarol:
Wow! Thanks @xbuffat!
I'll merge in a dev branch and then into the master

PR #50, opened 2021-09-03
Make BeamInteraction thick

anabramo:
Length is now a kwarg and a new attribute is added.

### DIFF
--- a/xtrack/beam_elements/beam_interaction.py
+++ b/xtrack/beam_elements/beam_interaction.py
@@ -4,8 +4,12 @@ from ..particles import Particles
 
 class BeamInteraction:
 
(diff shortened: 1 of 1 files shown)


giadarol:
@anabramo I propose the following implementation, which allows an explicit control of the ```isthick``` flag
```python
def __init__(self, name, interaction_process, length=0, isthick=None):
    self.name = name
    self.interaction_process = interaction_process

    self.length = length
    
    if isthick is None:
        isthick = True if length > 0 else False
   self.isthick = isthick
```

anabramo:
@giadarol I've implemented the change you suggest, the pr should be updated.

PR #63, opened 2021-09-29
Adds DriftExact from pysixtrack / sixtracklib

martinschwinzerl:
Adds a DriftExact beam-element. This is required to perform simulations with lattices where the parabolic approximation is not upheld (e.g., SIS100 at GSI, for example).


giadarol:
To be reopened if completed

PR #121, opened 2022-02-16
Change element name behaviour in new xtrack.Line

anabramo:


giadarol:
Included in version 0.9.0

PR #147, opened 2022-03-24
v0.10.1

giadarol:
**Changes:**
 - Fix in import RF cavity from MAD-X when the harmonic number is provided.
 - Fix in wire import from MAD-X.
 - Example with Wakefield on CPU together with lattice tracking and spacecharge on GPU.
 - Introduced test for mad import for several elements.
 - Other minor fixes.


xtrack/loader_mad.py
```diff
@@ -213,11 +217,19 @@ def madx_sequence_to_xtrack_line(
                     wire_xma     = ee.xma[0],
                     wire_yma     = ee.yma[0]
                 )
+                line.element_dict[eename] = newele
```
giadarol, line 220: Issue with wire import was here

PR #156, opened 2022-03-31
Matrix element

xbuffat:
Implementation of the 'Matrix' element from MAD-X including the loading from a sequence


tests/test_import_madx.py
```diff
@@ -177,3 +176,10 @@ def test_mad_element_import():
     assert line['w'].wire_L_int == 2
     assert line['w'].wire_xma == 1e-3
     assert line['w'].wire_yma == 2e-3
+
+    assert isinstance(line['mat0'],xt.FirstOrderTaylorMap)
+    assert line.get_s_position('mat0') == 2
+    assert np.allclose(line['mat0'].m0,matrix_m0)
+    assert np.allclose(line['mat0'].m1,matrix_m1)
```
giadarol, line 183: Default tolerances of np.allclose are quite large. Better to specify appropriate atol and troll values

giadarol (changes requested):

xbuffat:
In the meantime I found a bug in the loader, I created a new pull request.

PR #166, opened 2022-04-23
Partial tracking with collective elements

freddieknets:
The following new features are implemented:

1) A stop element can be specified with `ele_stop` (in which case `num_elements` has to be `None`).
Tracking will then go from `ele_start` (defaulting to 0) until `ele_stop`, potentially over several turns.
If `ele_stop` is before `ele_start` we pass `s=0`. This counts as one turn (though inside the `particles`
object the `at_turn` will be at 1 already).

The logic is hence that, when `ele_stop` is used, the last turn will be incomplete.

Examples:
  a) `ele_start=IP8`, `ele_stop=IP2`, `num_turns=1`
  This means we start at IP8, cross `s=0`, and continue until IP2.
  This is less than a full turn.
  Particles that died between IP8 and IP1 will be `at_turn=0`.
  Particles that died between IP1 and IP2 will be `at_turn=1`.

  b) `ele_start=IP8`, `ele_stop=IP2`, `num_turns=2`
  This means we start at IP8, make a full turn back to IP8, and then continue until IP2.
  This is more than one turn, but less than two.
  Particles that died between IP8 and IP1 will be `at_turn=0`.
  Particles that died between IP1 and IP1 (next turn) will be `at_turn=1`.
  Particles that died between IP1 (next turn) and IP2 (next turn) will be `at_turn=2`.

2) Both `ele_start` and `ele_stop` can be given by either the element index, or the element name.

3) Partial tracking is fully implemented for tracking with collective elements as well.

4) Small fix of an existing typo inside `_track_with_collective` (line 722-723):

  if _need_clean_active_lost_state:
    particles._num_active_particle = -1

is replaced with

  if _need_clean_active_lost_state:
    particles._num_active_particles = -1

### DIFF
--- a/xtrack/tracker.py
+++ b/xtrack/tracker.py
@@ -633,10 +633,23 @@ class Tracker:
         self,
         particles,
         ele_start=None,
+        ele_stop=None,
         num_elements=None,
         num_turns=1,
         turn_by_turn_monitor=None,
     ):
+    # If num_elements is not None, we track from ele_start until (ele_start + num_elements).
+    # In that case, ele_stop has to be None, and num_turns has to be 1.
+    #
+    # On the other hand, if ele_stop is not None, we track from ele_start until ele_stop.
+    # In that case, num_elements has to be None, but num_turns can be anything. If num_turns
+    # is 1, we just track between the two elements, even if that means passing over s=0
+    # (internally the end particles will be already at turn 2 though).
+    #
+    # Example: ele_start=IP8, ele_stop=IP2, num_turns=2 means we start at IP8, make a
+    # full turn back to IP8, and then continue until IP2.
+    #
+    # The logic is that, when ele_stop is used, the last turn will be incomplete.
 
         if particles.start_tracking_at_element >= 0:
             assert ele_start is None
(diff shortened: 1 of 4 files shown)


fsoubelet:
> Particles that died between IP1 (next turn) and IP8 (next turn) will be `at_turn=2`.

Just wanted to point out this should be `between IP1 (next turn) and IP2 (next turn)`. Otherwise very clear PR 👍

freddieknets:
Ah absolutely true!
Thanks for spotting it.

freddieknets:
Also, I just realised that this PR includes the changes from my previous PR #164, as I had already merged them in my main branch locally. But it's easy to disentangle them, as the previous changes are in a different file (loader_sixtrack.py).

xtrack/tracker.py
```diff
@@ -720,9 +757,20 @@ def _track_with_collective(
...
 
                 if _need_clean_active_lost_state:
-                    particles._num_active_particle = -1
+                    particles._num_active_particles = -1
                     particles._num_lost_particles = -1
 
+                if particles._num_active_particles == 0 or len(particles.state) == 0:
+                    stop_tracking = True
+
+                # Break from loop over particles if all are lost or if stop element reached
```
giadarol, line 766: ```loop over particles``` should be ```loop over parts```. I can correct it on the fly when I merge

freddieknets:
There were a few loose ends, ugly code, and the partial tracking was note done correctly in all cases.

These things are addressed in the last commit ([723796d](https://github.com/xsuite/xtrack/pull/166/commits/723796dd31367c22b00fe09bc2b17f77e33c0506)). Furthermore:

1. Thorough tests are added, to check every possible combination of input arguments
2. The argument `num_elements` can now be larger than the number of elements in the lattice; it will then overflow in the next turn. To make this logically sound, `num_elements` can no longer be used together with `num_turns`. The idea is to use either `ele_start` + `num_elements`, or `ele_start` + `ele_stop` + `num_turns`.

This implies the following possible usages:

- `track()`: track one full turn
- `track(num_turns=n)`: track `n` full turns
- `track(ele_start=start)`: track from the element `start` to the end of the lattice (less than one turn)
- `track(ele_start=start, num_turns=n)`: track from the element `start` to the end of the lattice, and add `n-1` full turns (less than `n` turns in total)
- `track(ele_stop=stop)`: track from the beginning of the lattice to the element `stop`(less than one turn)
- `track(ele_stop=stop, num_turns=n)`: track `n-1` full turns, and continue tracking to the element `stop` (less than `n` turns in total)
- `track(ele_start=start, ele_stop=stop)`: track from the element `start` to the element `stop`(less than one turn)
- `track(ele_start=start, ele_stop=stop, num_turns=n)`: track from the element `start` to the end of the lattice, continue by tracking `n-2` full turns, and continue tracking to the element `stop` (less than `n` turns in total)
- `track(ele_start=start, num_elements=n)`: track from the element `start` to `start+num_elements`. If the latter is more than the number of turns, tracking overflows in the next turn. There is no limit on the number of turns it can overflow into.

PR #179, opened 2022-05-04
Fixed language recognition.

freddieknets:
Very trivial PR - just adjusted the auto-language recognition such that the repo is not listed as 'Roff'


giadarol:
Does not change the code, I'll merge it right away

PR #180, opened 2022-05-06
Update synrad_spectrum.h

rdemaria:
q0^2 goes to the numerator

image

image

image

Needs to be update using `mratio` and `qratio`


rdemaria:
Could also be rewritten using
image

PR #219, opened 2022-10-08
Feature/survey

pbelange:
Done. Should maybe verify that the order of the elements `#s/#e/_end_point/ip3 markers` in twiss.mirror is the good one since I had to reshuffle them in survey.mirror


giadarol:
> Done. Should maybe verify that the order of the elements `#s/#e/_end_point/ip3 markers` in twiss.mirror is the good one since I had to reshuffle them in survey.mirror

In fact I would not move the markers. For me the sequence is mirrored and therefore the markers are flipped

pbelange:
but you still want the sequence to start at `ip3` at least? no?

giadarol (commented):
I note down here a few aspects that I am changing. To be discussed with @pbelange

PR #227, opened 2022-10-26
performance: special di- and quadrupoles, add defines for disabling parts of tracker

szymonlopaciuk:
## Description

Create optimised versions of dipoles and quadrupoles. Add compile-time flags for disabling the EBE monitor and active check in tracker.

Part of a solution for xsuite/xsuite#164.

## Checklist

Mandatory: 

Optional:

### DIFF
--- a/tests/test_line.py
+++ b/tests/test_line.py
@@ -272,3 +272,32 @@ def test_from_dict_current():
     assert d1.length == 4
 
     assert d2 is d1
+
+
+def test_optimize_multipoles():
+    elements = {
+        'q1': xt.Multipole(knl=[0, 1], length=0),
+        'q2': xt.Multipole(knl=[1, 2], length=0),
+        'q3': xt.Multipole(knl=[0, 1], length=1),
+        'q4': xt.Multipole(knl=[0, 1], ksl=[1, 2], length=0),
+        'd1': xt.Multipole(knl=[1], hxl=0.0, length=2),
+        'd2': xt.Multipole(knl=[1], hxl=0.1, length=0),
+        'd3': xt.Multipole(knl=[1], hyl=1, length=2),
+        'd4': xt.Multipole(knl=[1], ksl=[3], length=2),
+    }
+
+    test_line = xt.Line(
+        elements=elements,
+        element_names=elements.keys(),
+    )
+
+    test_line.optimize_dipoles()
+    test_line.optimize_quadrupoles()
+
+    def keys_for_elements_of_type(type_):
+        for kk, vv in test_line.element_dict.items():
+            if isinstance(vv, type_):
+                yield kk
+
+    assert set(keys_for_elements_of_type(xt.FastDipole)) == {'d1', 'd2'}
+    assert set(keys_for_elements_of_type(xt.FastQuadrupole)) == {'q1'}
--- a/tests/test_line.py
+++ b/tests/test_line.py
@@ -4,7 +4,6 @@
 # ######################################### #
 
 import numpy as np
-import pytest
 import xtrack as xt
 
 def test_simplification_methods():
@@ -291,13 +290,13 @@ def test_optimize_multipoles():
         element_names=elements.keys(),
     )
 
-    test_line.optimize_dipoles()
-    test_line.optimize_quadrupoles()
+    test_line.use_simple_bends()
+    test_line.use_simple_quadrupoles()
 
-    def keys_for_elements_of_type(type_):
-        for kk, vv in test_line.element_dict.items():
-            if isinstance(vv, type_):
-                yield kk
-
-    assert set(keys_for_elements_of_type(xt.FastDipole)) == {'d1', 'd2'}
-    assert set(keys_for_elements_of_type(xt.FastQuadrupole)) == {'q1'}
+    for nn in test_line.element_names:
+        if nn in ('d1', 'd2'):
+            assert type(test_line.element_dict[nn]) is xt.SimpleThinBend
+        elif nn == 'q1':
+            assert type(test_line.element_dict[nn]) is xt.SimpleThinQuadrupole
+        else:
+            assert type(test_line.element_dict[nn]) is xt.Multipole
--- a/tests/test_line.py
+++ b/tests/test_line.py
@@ -4,6 +4,7 @@
 # ######################################### #
 
 import numpy as np
+import xobjects as xo
 import xtrack as xt
 
 def test_simplification_methods():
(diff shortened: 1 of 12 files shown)


tests/test_line.py
```diff
@@ -272,3 +272,32 @@ def test_from_dict_current():
     assert d1.length == 4
```
giadarol, line 272: Somewhere we should also test that the behaviour of the knobs (tracker.vars) is preserved, as done in this test:
https://github.com/xsuite/xtrack/blob/c39c22728e43a582f4405e7220ebcb381399f1af/tests/test_vs_madx.py#L455

Also it would be good to add a test for the tracker with all speedup modifications (no rad, simple quads, etc.)

tests/test_multisetter.py
```diff
@@ -62,4 +63,64 @@ def test_custom_setter_scalar():
         assert np.all(ctx2np(qf_setter.get_values()) == np.array([10., 100., 1000.]))
         assert line['qf1'].length == 10.
         assert line['qf2'].length == 100.
-        assert line['qf3'].length == 1000.
\ No newline at end of file
+        assert line['qf3'].length == 1000.
+
+
+def test_custom_setter_fast_quadrupole():
```
giadarol, line 69: I would rename FastQuadrupole in to SimpleThinQuadrupole and FastDipole into SimpleThinBend.
Of course all names in tests, examples, etc. need to follow

tests/test_line.py
```diff
@@ -272,3 +272,32 @@ def test_from_dict_current():
...
+    test_line.optimize_dipoles()
+    test_line.optimize_quadrupoles()
+
+    def keys_for_elements_of_type(type_):
+        for kk, vv in test_line.element_dict.items():
+            if isinstance(vv, type_):
+                yield kk
+
+    assert set(keys_for_elements_of_type(xt.FastDipole)) == {'d1', 'd2'}
+    assert set(keys_for_elements_of_type(xt.FastQuadrupole)) == {'q1'}
```
giadarol, line 303: I would prefer something more explicit:

```python
for nn in line.element_names:
    if nn in ('d1', 'd2'):
        assert type(line[nn] is xt.SimpleThinBend)
    elif nn in ('q1'):
        assert type(line[nn] is xt.SimpleThinQuadrupole)
    else:
        assert type(line[nn] is xt.Multipole)

```

tests/test_line.py
```diff
@@ -272,3 +272,32 @@ def test_from_dict_current():
     assert d1.length == 4
 
     assert d2 is d1
+
+
+def test_optimize_multipoles():
```
giadarol, line 277: I would rename FastQuadrupole in to **SimpleThinQuadrupole** and FastDipole into **SimpleThinBend**.
Of course all names in tests, examples, etc. need to follow

xtrack/beam_elements/elements.py
```diff
@@ -336,6 +336,81 @@ def get_backtrack_element(self, _context=None, _buffer=None, _offset=None):
...
+
+            - knl [m^-n, array]: Normalized integrated strength of the normal components.
+            - hxl [rad]: Rotation angle of the reference trajectory in the horizontal plane.
+            - length [m]: Length of the originating thick multipole.
+
+    """
+
+    _xofields={
+        'knl': xo.Float64[1],
+        'hxl': xo.Float64,
```
giadarol, line 385: We can assume hxl = knl[0]. Maybe it gives us a visible speedup?

xtrack/line.py
```diff
@@ -638,6 +638,41 @@ def merge_consecutive_multipoles(self, inplace=True):
         self.element_names = newline.element_names
         return self
 
+    def optimize_quadrupoles(self):
```
giadarol, line 641: We could call it `use_simple_quadrupoles`

xtrack/line.py
```diff
@@ -638,6 +638,41 @@ def merge_consecutive_multipoles(self, inplace=True):
...
+                    not np.any(el.ksl) and
+                    not el.hxl and
+                    not el.hyl)
+
+        for name, element in self.element_dict.items():
+            if is_simple_quadrupole(element):
+                fast_quad = beam_elements.FastQuadrupole(knl=element.knl)
+                self.element_dict[name] = fast_quad
+
+    def optimize_dipoles(self):
```
giadarol, line 659: we could call it use simple bend

xtrack/line.py
```diff
@@ -638,6 +638,41 @@ def merge_consecutive_multipoles(self, inplace=True):
...
+        def is_simple_dipole(el):
+            if not isinstance(el, beam_elements.Multipole):
+                return False
+            return el.order == 0 and not np.any(el.ksl) and not el.hyl
+
+        for name, element in self.element_dict.items():
+            if is_simple_dipole(element):
+                fast_di = beam_elements.FastDipole(
+                    knl=element.knl,
+                    hxl=element.hxl,
```
giadarol, line 671: If assume hxl = knl[0] this should be added to the checks above

tests/test_line.py
giadarol, line 277: We should check that this test works also if the elements are on GPU

xtrack/tracker.py
```diff
@@ -655,11 +657,13 @@ def _build_kernel(self, save_source_as, compile):
         src_lines.append(
             """
                         } //switch
+                    #ifndef SKIP_ACTIVE_CHECK_AND_SWAPS
```
giadarol, line 660: This is a bit dangerous (I put it in only to see)... How much do we lose if we don't skip it?

giadarol (changes requested):

xtrack/beam_elements/elements.py
szymonlopaciuk: I ran it a few times and it seems that it is consistently faster, but only by like 0.3% on average on the simulation you gave me. There is a difference in the number of elements that can be converted too, and it's similarly insignificant.

xtrack/tracker.py
szymonlopaciuk: Actually a lot... It totals a third of all the savings we make in this PR (on the same example simulation as above).

This is my summary:
| scenario | time/particle/turn [us] |
|---|---|
| everything on + faster Faddeeva | 54.78 (+37%) |
| everything in this PR on, no Faddeeva | 65.65 (+24%) |
| no `SKIP_ACTIVE_CHECK_AND_SWAPS`, ditto | 80.18 (+7%) |
| everything in this PR off, ditto | 86.58 |

PR #233, opened 2022-11-04
Added sps ion json line with particles

sailew:
Added script to prepare line identical to existing `000_make_line.py` but added particle to be stored, for easier ion tests. Now `line_with_particle.json` also contains a particle object.


giadarol:
I prefer to add the reference particle directly in ```000_make_line.py``` and avoid the duplication of the script.
I can do it myself.

sailew:
> I prefer to add the reference particle directly in `000_make_line.py` and avoid the duplication of the script. I can do it myself.

Okay makes sense! Should I remove the pull request or keep it for now?

PR #253, opened 2022-11-21
Add defaults for LimitRect aperture

eltos:
## Description

Currently the aperture of a `LimitRect` defaults to `0` if not specified, meaning that e.g.
```
xt.LimitRect(min_x=-0.1, max_x=0.1)
```
is a zero-size aperture in y and will block all particles, which is counter-intuitive.

This PR sets default limits to infinity, such that only the limits actually specified have an effect, e.g.
```
xt.LimitRect(min_x=-0.1)
```
is an aperture limitation only in `x<0` and unbound in `x>0` as well as `y`.


giadarol:
I see a few things to do if we want to go in this direction:
 - Adapt all the other aperture elements to be  "open" with default initialisation parameters
 - Make sure that the infs are correctly passed between python and c on all CPU and GPU contexts (tests to be added to the test suite)
 - Make sure that ```to_json``` and ```from_json``` behave correctly when infs are present.

Alternatively we could just avoid default parameter altogether and raise an exception if the user misses a parameter.

Another alternative to achieve the same behavior as you suggest is to just use large numbers (e.g. 9999999999 meters) as default parameters. Not very elegant but robust.

Happy to hear your thoughts

eltos:
To my knowledge, infs and nans are nothing special, they are just part of the floating point standard (IEEE 754). I have used infs for min_y and max_y without problems on a GPU with the pyopencl context.
Here are a few links to reassure:
- [Opencl specs](https://registry.khronos.org/OpenCL/specs/opencl-1.2.pdf) section 7.2
- [CUDA documentation](https://docs.nvidia.com/cuda/floating-point/index.html#formats) section 2.1
- [CuPy documentation](https://docs.cupy.dev/_/downloads/en/latest/pdf/) search for `cupy.inf`
- [Python docs on json](https://docs.python.org/3/library/json.html#json.dump) search for allow_nan (default: True)

Of course adding tests is always a good idea and if you prefere using 1e38 is also fine for real world problems.

giadarol:
I agree that infs should indeed work.
I would prefer to use a large number, for two reasons:
 - It spares us some testing
 - It works also for LimitEllipse and LimitRectEllipse, where the inf would not really work in the formula

@eltos if you agree, would you be available to make these modifications?

eltos:
> I would still prefer to use a large number

Fine for me

giadarol (approved):

giadarol:
Merged in v0.23.5 release branch

giadarol:
@eltos Thanks a lot!

xtrack/beam_elements/apertures.py
```diff
@@ -57,12 +86,18 @@ def to_dict(self):
...
+        """An elliptical aperture
+        
+        Args:
+            a (float): Horizontal semi-axis of ellipse in m
+            b (float): Vertical semi-axis of ellipse in m
+        """
 
         if a is None and a_squ is None:
-            a = 1.
+            raise ValueError("Horizontal semi-axis of ellipse not specified.")
```
giadarol, line 97: Why don't we use "UNLIMITED" also here?

eltos: In my opinion, it is pointless to have an elliptic aperture with one dimension unlimited, since there is nothing elliptical about this aperture anymore, it's just a slit. One should use a LimitRect instead, which is faster in tracking, clearer in code, and - in contrast to the "ellipse" - numerically stable[^1]. If one of the semi-axes are not given, it's likely the user forgot it.

What we could do, instead of raising an Exception, is to default to a circular aperture (i.e. a=b), if you want?

[^1]: The ellipse with one very large semi-axis is numerically unstable, since in the code `double const temp = x*x*b_squ + y*y*a_squ;`, the small float being added to the large float will be lost due to insufficient precision.

giadarol: I will merge it with UNLIMITED for now (I reduced UNLIMITED to 1e10). We can change it later

PR #256, opened 2022-11-22
generate and reuse compiled track kernels

szymonlopaciuk:
## Description

This adds a functionality of precompiling tracking kernels, so they can be reused between sessions.

The track kernels can be precompiled using `xtrack.prebuild_kernels.precompile_kernels`. Additionally the module contains the descriptions of the kernels to be generated. When building the tracker, in `xtrack.Tracker._build_tracker` an additional parameter `built_ffi_module_name` can be passed, which will try to load the kernel from the module of the given name, and failing that will compile the kernel and save it as that module.

A new value for the parameter `compile` in `Tracker._build_tracker` is introduced: it can now be either bool or 'force'. See the docstring of the method for more details.

Needs xsuite/xobjects#77.

Closes xsuite/xsuite#224.

## Checklist

Mandatory: 

Optional:

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -37,9 +37,10 @@ setup(
             "Source Code": "https://github.com/xsuite/xtrack",
         },
     packages=find_packages(),
-    ext_modules = extensions,
+    ext_modules=extensions,
     include_package_data=True,
     install_requires=[
+        'cffi>=1.0.0',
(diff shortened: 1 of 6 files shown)


szymonlopaciuk:
Obsolete

PR #271, opened 2022-12-13
Fixed non-symplecticity bug in LinearTransferMatrix

pkruyt:
## Description

1. Added momentum dispersion
2. Fixed symplecticity by including transformation as described by Eq 3.41 in Y. Cai, Symplectic maps and chromatic optics in particle accelerators, NIM-A  797, 2015, Pages 172-181 (https://doi.org/10.1016/j.nima.2015.06.059). The variable 'l' in the paper is changed to xpart's xi (xi = -l)' and transformed to zeta (zeta = xi / rvv)  
4. Added the relevant tests
Closes # .

## Checklist

Mandatory: 

Optional:


xbuffat:
Ok for me!

lineartransfermatrix.h
```diff
@@ -0,0 +1,219 @@
+// copyright ############################### //
```
giadarol, line 1: I believe this file is a duplicate and should be removed

xtrack/beam_elements/elements.py
```diff
@@ -885,3 +895,99 @@ def __init__(self, length = 0.0, m0 = None, m1 = None,radiation_flag=0,**nargs):
         super().__init__(**nargs)
 
 
+class IonLaserIP(BeamElement):
```
giadarol, line 898: I believe this element was added by mistake and should be removed (otherwise tests should be added)

xtrack/beam_elements/elements_src/lineartransfermatrix.h
```diff
@@ -61,12 +67,15 @@ void LinearTransferMatrix_track_local_particle(LinearTransferMatrixData el, Loca
     double new_px = LocalParticle_get_px(part);
     double new_py = LocalParticle_get_py(part);
     double delta = LocalParticle_get_delta(part);
+    double zeta_no_disp = LocalParticle_get_zeta(part);
 
     // removing dispersion and close orbit
     new_x -= disp_x_0 * delta + x_ref_0;
-    new_px -= px_ref_0;
+    new_px -= disp_px_0 * delta + px_ref_0;
```
giadarol, line 74: delta and zeta are not canonically conjugate. Using a symplectic matrix with these coordinates results in a non-symplectic map

giadarol (changes requested):
Thank you very much for this PR. See below some comments.

In general, the publication you quote uses a different couple of conjugate variables for the longitudinal plane (delta, L=beta c t) compared to xsuite that uses (zeta=-beta0 c t, pzeta - see xsuite physics manual for the definitions). Note the swap (in the paper the momentum is L while for us it is pzeta). Moreover the paper assumes a drift in the longitudinal plane while in the xsuite element we have a rotation.

Could you please prepare a short latex document (on overleaf for example) where you adapt the derivation from the paper to our case?

This will allow checking that the proposed modification is correct.

Such a document could also be added as a section of the xsuite physics manual.

xbuffat:
Indeed in the code we implemented a modified version of the matrix in the initial reference which was based on a different definition of the longitudinal variable. I changed the reference in the description to another one with the right convention. I would propose to merge this, as anyway I need to write the physics manual for this element. I will also integrate this part them.

xbuffat:
Actually it is still not exactly the same (in the non ultrarelativistic case). Let's redo the math then before we merge...

xbuffat:
I checked the math and added the missing beta ratios (rvv). I'll include the math when writing the doc for the entire element which is still missing.
There is an error that is generated by the energy_ref_increment that is linked to the fact that rvv is not properly updated in ducktrack (It was already there before, but remained unnoticed since in the tests rvv was not used). I suggest to fix this in a separate issue, for the moment I set it to 0 in the tests.

PR #278, opened 2022-12-17
Release/0.25.0

giadarol:
**Changes:**
 - Possibility to use prebuilt kernels instead of compiling just in time.
 - In ```tracker.compensate_radiation_energy_loss``` added ```record_iterations``` flag.
 - Added ```Marker``` element type and modified mad loader to use markers instead of zero-length drifts.
 - In line optimizers, added possibility of preserving all markers or selected ones.
 - Added ```Line.from_sequence``` constructor.
 - Added some missing backtracking elements.
 - Added "exact" drift mode.


giadarol:
Merged

PR #279, opened 2023-01-10
Exciter beam element

eltos:
## Description

Adds a new `Exciter` beam element which models a transverse exciter as a time-dependent thin multipole.

This is motivated by simulating excitation for RF Knock-Out extraction, but can be used for any time-dependent transverse multipole, e.g. excitation for tune measurements, power supply ripples, tune wobbling, etc.
Time-dependence comes via an array of samples, which allows for slow and fast signals of arbitrary frequency spectrum. The sampling rate can be adjusted according to the needs. Also, the particle arrival time is taken into account. More details are given in the docstrings:

    Beam element modeling a transverse exciter as a time-dependent thin multipole.
    
    The given multipole components (knl and ksl) are scaled according to an array of
    samples, allowing for arbitrary time dependence:
        
        knl(t) = knl * samples[i]

    It is *not* assumed that the variations are slow compared to the revolution frequency
    and the particle arrival time is taken into account when determining the sample index:
    
        i = sampling * ( ( at_turn - start_turn ) / f_rev - zeta / beta0 / c0 )
    
    where zeta=(s-beta0*c0*t) is the longitudinal coordinate of the particle, beta0 the
    relativistic beta factor of the particle, c0 is the speed of light, at_turn is the
    current turn number, f_rev is the revolution frequency, and sampling is the sampling
    frequency. The excitation starts with the first sample when the reference particle 
    arrives at the element in start_turn.

    You can think of the samples as a waveform recorded at the sampling frequency.
    For example, to compute samples for a sinusoidal excitation with frequency f_ex one
    could do: samples[i] = np.sin(2*np.pi*f_ex*i/sampling).    
    
    Notes:
        - This is not to be confused with an RFMultipole, which inherits the characteristics
          of an RFCavity and whose oscillation is therefore with respect to the reference
          particle. While the frequency of the RFMultipole is therefore restricted to
          harmonics of the revolution frequency, the exciter allows for arbitrary frequencies.
        - This is also not to be confused with an ACDipole, for which the oscillation is 
          assumed to be slow compared to the revolution frequency and the kick is the same 
          for all particles independent of their longitudinal coordinate.
        
        
    Parameters:    
        - knl (float array): Normalized integrated strength of the normal components. Unit: m^-n (n=0,1,2,...).
        - ksl (float array): Normalized integrated strength of the skew components. Unit: m^-n (n=0,1,2,...).
        - samples (float array): Samples of excitation strength to scale knl and ksl as function of time.
        - sampling (float): Sampling frequency in Hz.
        - frev (float): Revolution frequency in Hz of circulating beam (used to relate turn number to sample index).
        - start_turn (int): Start turn of excitation.
        - nsamples (int): Number of samples. Pass this instead of samples to reserve memory for later initialisation.
        - order (int): Multipole order (readonly), i.e. largest n with non-zero knl or ksl.
        

**Example:**
```python

fs = 10e6 # sampling frequency in Hz

# load waveform into memory
signal = np.copy(np.memmap("signal.10MHz.float32", np.float32))

# alternatively compute samples on the fly, for example a simple sine at 500 kHz ...
t = np.arange(1000)/fs
f_ex = 5e5 # excitation frequency in Hz
signal = np.sin(2*np.pi*f_ex*t)
# ... or a sweep from 500 to 800 kHz
f_ex_1 = 8e5
signal = scipy.signal.chirp(t, f_ex, t[-1], f_ex_1)

# create the exciter
frev = 1e6 # revolution frequency in Hz
k0l = 0.1 # this is scaled by the signal 
exciter = Exciter(samples=signal, sampling=fs, frev=frev,
                  start_turn=0, knl=[k0l])

# add it to the line
line.insert_element(
    index=...,
    name=...,
    element=exciter,
)

```

## Checklist

~~*This is considered a draft until tests are added*~~

Mandatory: 

Optional:


ab-gh:
I'm currently testing this branch and working on an implementation of the exciter, but it looks like this PR has fallen behind the main branch and is no longer compatible - especially with the changes to random number generation.

Would it be possible to update this PR/fork with the latest changes in xsuite/xtrack main branch ?

giadarol:
If this is used it makes sense to finalise the PR and merge it in the production version.

@eltos I see that this is still a draft, is there anything you would like to change? Apart from some variable names that I would like to change, for me this is good to go.

eltos:
I changed the PR status from "draft" to "ready for review" last week after adding the unit tests. It looks like I forgot to remove the comment in the description, which I did now.

**From my point of view it is ready to be reviewed and merged.**

@ab-gh The branch was indeed outdated when we talked last time. I have now synced it and fixed an import. The Exciter related unit tests are passing (see screenshot below) so it should be fine now? I am not sure what you mean with the random number generator, since it's not used by the Exciter. Could you check again with the most recent version(s) and - in case there are still issue - provide a bit more detail what you are testing and working on exactly?

@giadarol please go ahead and rename the variables as you prefer. I have checked the "Allow edits by maintainers" options so you should be able to add commits to the PR directly.


grafik


/cc @eccortes95

ab-gh:
Thanks for updating the branch - the random number generator issue was just caused by outdated code elsewhere in the branch which broke compatibility with Xfields - but its all working now, thank you!

giadarol:
For simplicity of testing (to have data aligned with 64 bits) I changed the type of ```samples``` from Float32 to Float64.
@eltos Is that a problem?

giadarol:
> [2966bcf](https://github.com/xsuite/xtrack/commit/2966bcf1007a140866d5147f87a4840c44334a67)
> Whats is the issue with a Float32 array on GPU contexts? I have been using 32 bit waveforms successfully on AMD GPUs with the opencl context for quite a while now. Could you elaborate a bit?

In fact there is no problem, it is just that the test for Element.copy needs to become a bit smarter. 

For pyopencl we try to facilitate SIMD alignment of the data by inserting some little empty space where needed. This means that in the presence of types smaller that 64 bit equality of two objects between CPU and GPU cannot be checked comparing the buffers bit to bit.

Since this can be limiting for you, I will go back to your implementation with the float32 and find a way to adapt the test.

giadarol:
> > [2966bcf](https://github.com/xsuite/xtrack/commit/2966bcf1007a140866d5147f87a4840c44334a67)
> > Whats is the issue with a Float32 array on GPU contexts? I have been using 32 bit waveforms successfully on AMD GPUs with the opencl context for quite a while now. Could you elaborate a bit?
> 
> In fact there is no problem, it is just that the test for Element.copy needs to become a bit smarter.
> 
> For pyopencl we try to facilitate SIMD alignment of the data by inserting some little empty space where needed. This means that in the presence of types smaller that 64 bit equality of two objects between CPU and GPU cannot be checked comparing the buffers bit to bit.
> 
> Since this can be limiting for you, I will go back to your implementation with the float32 and find a way to adapt the test.

The change back to float will be coming soon. 

Note that a way of overcoming such a limitation is to update the samples content every n turns from python. The overhead of going back to python and calling Tracker.track again is typically negligible (we are careful to keep this fast) . We do it  even once per turn for a relatively big machine without suffering much in speed.

eltos:
@giadarol Thank you very much.
I think testing supporting for float32 will also be beneficial for other cases in the long term.
Indeed, updating the samples is a way to overcome this limitation. Due to the fact that the Exciter is time (and not turn) based, this is not straight forward and requires ensuring sufficient overlap of samples to cover all particle arrival times (especially for coasting beam studies). But it is definitely possible.

PR #280, opened 2023-01-12
Initialize the _beamstrahlung_model

anabramo:
Initialize the _beamstrahlung_model variable to avoid situations where it is used/modified without being initialized via tracker.configure_radiation

## Description
Following the fixes in 0.25.1, this variable beamstrahlung_model is copied in _build_auxiliary_tracker_with_extra_markers, but this can be used without running tracker.configure_radiation() first, for example in the build_particles method. 

Closes [Xsuite issue #269](https://github.com/xsuite/xsuite/issues/269)

## Checklist

Mandatory: 

Optional:


anabramo:
Also added an optimisation for the tapering. The access to the the multipole elements went via line.__getitem__, checking that the name is in the line via an O(n) 'x in list' check, which scales terribly for the FCC and dominates the run time. As the element names are collected from the line for the tapering, the check is not needed and accessing via line.element_dict directly completely removes the bottleneck. Should check for other internal usages of line.__getitem__ that may suffer the same issue.

PR #295, opened 2023-02-03
Add LastTurnsMonitor

eltos:
## Description

Adds a new monitor to record particle data in the last turns before they are lost.
This is motivated by studies of slow extraction, but I think it can be useful for other cases where beam dynamics shortly before particle loss are of interest. So I wanted to share it.

The idea is to use a rolling buffer instead of saving *all* the turns. This saves a lot of memory resources, especially when the interest lies only in the last few turns anyway. When accessing the recorded data via the monitors attributes, the buffers are automatically rolled according to their offsets, such that the index `[:,-1]` corresponds to the last turn for each of the particles, respectively. It is also possible to store only every 2nd, 3rd, etc. turn.
 
```
Monitor to save particle data in last turns before respective particle loss
        
The monitor provides the following data as 2D array of shape (num_particles, n_last_turns),
where the first index corresponds to the particle_id in particle_id_range
and the second index corresponds to the turn (or every_n_turns) before the respective particle is lost:
`particle_id`, `at_turn`, `x`, `px`, `y`, `py`, `delta`, `zeta`

Args:
    n_last_turns (int): Amount of turns to store before particle loss.
    particle_id_range (tuple): Range of particle ids to monitor (start, stop).
    num_particles (int, optional): Number of particles. Equal to passing particle_id_range=(0, num_particles).
    every_n_turns (int, optional): Save only every n-th turn, i.e. turn numbers which are a multiples of this.
        Because `n_last_turns` defines the amount of turns to store (and not the range), the data will cover turn
        numbers up to `n_last_turns*every_n_turns` turns before particle loss.
        
Example:
    monitor = LastTurnsMonitor(n_last_turns=5, particle_id_range=(1, 5))
    monitor.at_turn[:,-1]  # last turn before loss of each particle, respectively
    monitor.x[3,-2]  # x coordinate in one but last turn of particle with id 4
        
```

The monitor uses 32 bit data types and only stores the 6D phase space coordinates to keep the memory footprint small.

I moved the file(s) to the `monitors` folder to keep the file structure overseeable when future monitor types are added.

## Checklist

Mandatory: 

Optional:


giadarol:
I like it very much, it can be very useful for many studies.

I would propose to rename the class to "LastTurnsMonitor" because actually this works also if the particle is not lost and in that case you get the last N turns before the end of the simulation.

giadarol (approved):

PR #297, opened 2023-02-06
Include chebychev polynomials 

phermes:
## Description

Introduced Chebychev polynomials to simulate residual fields of arbitrary shape acting on the beam core. Following the implementation in SixTrack. By default switched off. Only if the user provides a Chebychev reference radius different from zero, the polynomials are read and used. Polynomials of up to an arbitrary order can be used.

## Checklist

Mandatory: 

- [x ] I have added tests to cover my changes
- [ x] All the tests are passing, including my new ones
- [ x] I described my changes in this PR description

Optional:


xtrack/beam_elements/elements.py
```diff
@@ -138,29 +144,47 @@ def __init__(self,  inner_radius = 1.,
                         voltage      = 0.,
                         residual_kick_x = 0,
                         residual_kick_y = 0,
-                        coefficients_polynomial = [0],
+                        coefficients_polynomial     = [0],
```
giadarol, line 147: Use tuple instead of list

xtrack/beam_elements/elements.py
```diff
@@ -138,29 +144,47 @@ def __init__(self,  inner_radius = 1.,
...
+                        chebyshev_max_order         =  0 ,
+                        chebyshev_coefficients      =  [0],
```
giadarol, line 149: Same

tests/test_elements.py
```diff
@@ -292,6 +292,67 @@ def test_elens(test_context):
...
 
+
+
+@for_all_test_contexts
+def test_elens_chebychev(test_context):
+
+    # read the input from file
+    coeffs = []
+
+    with open('coeffs_gunBend_nEq18.txt') as f:
```
giadarol, line 303: Path needs to be generalised

giadarol (approved):

PR #301, opened 2023-02-14
Update random generator implementation.

freddieknets:
## Description

This PR is dependent on PRs xsuite/xpart#63, xsuite/xfields#75, and xsuite/xsuite#290.

1. Created random generator `xobject`s, one per distribution: `RandomUniform`, `RandomExponential`, `RandomNormal`, and `RandomRutherford`. The latter stores state variables from its distribution. Renamed all random sampling functions (`RandomUniform_generate` etc instead of `LocalParticle_generate_random_double`). This has two advantages:

    - a `BeamElement` can use the C random sampling functions just by adding `_depends_on = [ RandomUniform]` in its class definition (instead of adding `base_rng.h` and `local_particle_rng.h` to `_extra_c_sources`), or by adding an instance of a random generator to its `xofields` (which is needed if access to the state variables is required). All other dependencies for the random generator are then automatically included. 
    - the sampling code can be accessed in python for testing (and in case collective elements would want to use it)

The performance of `RandomRutherford` (used in xcoll) is very low, but on par with the original FORTRAN code.

2. Adapted  tests and examples where needed.

3. For the geometry `BeamElement`s (like `Drift`s and `Rotation`s): took out the effect on the single particles, such that those functions can be called from other C code (like in `xcoll`).

4. Added macros for particle lost state codes, and made random sampling fail if seeds are not set (by killing the particle).

Still missing:
- [x] Exposing `RandomRutherfordData_set` to python
- [x] Last test in test_element_internal_record.py fails but not clear why.

## Checklist

Mandatory: 

Optional:


freddieknets:
Moved most of the new implementation to xtrack, with only the core bitwise rng (base_rng.h) remaining in xpart (xsuite/xpart#63). Adapted this PR description to correctly describe the changes.

giadarol (approved):

PR #306, opened 2023-02-23
Bugfix/madx misalignments

kparasch:
## Description
Order of dx/dy and dpsi  when importing from madx error tables was wrong.
Added test to check, and reversed the order.

Fixes xsuite Issue # 302 (https://github.com/xsuite/xsuite/issues/302)

tests/test_prebuild_kernels.py does not pass, but was also not passing before.

## Checklist

Mandatory: 

- [] All the tests are passing, including my new ones

Optional:

### DIFF
--- a/tests/test_madloader.py
+++ b/tests/test_madloader.py
@@ -242,9 +242,6 @@ def test_shift_tilt_order():
             lrad=1;
     seq: sequence, l=1;
     elm1:elm, at=1;
-    !mk:marker, at=0.1;
-    !elm2:elm, at=0.5;
-    !elm3:elm, at=0.5;
     endsequence;
     beam;
     use,sequence=seq;
--- a/tests/test_madloader.py
+++ b/tests/test_madloader.py
@@ -229,6 +229,60 @@ def test_tilt_shift_and_errors():
         line=list(ml.iter_elements())
 
 
+def test_shift_tilt_order():
+    mad = Madx()
+    
+    src="""
+    k1=1.;
+    elm: multipole,
+            knl:={0.0,-k1},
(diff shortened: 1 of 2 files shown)


rdemaria:
Please make sure not to break the ealign logic.

kparasch:
Doesn't work if thin multipole has angle...

rdemaria:
> Doesn't work if thin multipole has angle...

Multipoles with angle are a bit special and we were considering changing the behaviour (as it is not documented yet). Have a look here for more details: 
https://github.com/MethodicalAcceleratorDesign/MAD-X/issues/911

kparasch:
> Multipoles with angle are a bit special and we were considering changing the behaviour (as it is not documented yet). Have a look here for more details: [MethodicalAcceleratorDesign/MAD-X#911](https://github.com/MethodicalAcceleratorDesign/MAD-X/issues/911)

Until (and if) behaviour is changed, should we replicate the track of mad-x?


A particle entering a multipole (knl,ksl = 0, angle=a), and with an ealign diplacement dx, exits with a shift in the x coordinate equal to "a * sin(a) + cos(a) - 1 + O(1.e-5)". Since there can't be a change in the actual position of the particle, this change must be happening in the reference orbit.

rdemaria:
The matter is a bit tricky. Thin elements do not exist, (si)xtrack elements are indeed integration steps. I am not sure it makes sense to rotate an integration step in the general case, but it can work in specific cases.

giadarol:
@rdemaria and @paraschou 
What is the state of this PR? Is it ready to be merged?

giadarol:
Closed by mistake

PR #310, opened 2023-03-04
Merge consecutive apertures

freddieknets:
## Description

1. Added a function `merge_consecutive_apertures()` to the `line` to collapse consecutive apertures: in the case of 3 or more consecutive apertures (only separated by `Marker`s or `Drift`s) that are exactly the same, all are removed except the first and the last
2. Expanded the other simplification methods (`remove_zero_length_drifts()`, `merge_consecutive_drifts()`,  `remove_markers()`, `remove_inactive_multipoles()`, and `merge_consecutive_multipoles()`): added a `keep` option to all, and implemented the `inplace=False` functionality
3. Added two tests for `merge_consecutive_apertures()` 
4. Expanded the existing tests for the other simplification methods, to take the new options for `keep` and `inplace` into account
5. Created two helper functions to the `line`: `_apertures_equal()` to check if two apertures are the same, and `_lines_equal()` which compares two lines for equality (this currently only works if all elements in the line have a `to_dict()` method, and raises a `NotImplementedError` otherwise)
6. Updated `check_aperture()` to use the `_is_aperture()` and `_allow_backtrack` helper functions. This improves code consistency, and avoids the need to update the code when new elements are added (like the `XRotation` and `YRotation`)

## Checklist

Mandatory: 

Optional:


xtrack/line.py
```diff
@@ -726,101 +768,177 @@ def get_s_position(self, at_elements=None, mode="upstream"):
...
-            names.append(nn)
+            if isinstance(ee, Marker) and nn not in keep:
+                continue
+            newline.append_element(ee, nn)
 
-        self.element_names = names
-        return self
+        if inplace:
+            self.element_names = newline.element_names
+            self.element_dict  = newline.element_dict
```
giadarol, line 788: I think rebinding element_dict breaks the xdeps expressions

xtrack/line.py
```diff
@@ -726,101 +768,177 @@ def get_s_position(self, at_elements=None, mode="upstream"):
...
                 ctx2np = ee._context.nparray_from_context_array
                 aux = ([ee.hxl, ee.hyl]
                         + list(ctx2np(ee.knl)) + list(ctx2np(ee.ksl)))
                 if np.sum(np.abs(np.array(aux))) == 0.0:
                     continue
             newline.append_element(ee, nn)
 
+        if inplace:
+            self.element_names = newline.element_names
+            self.element_dict  = newline.element_dict
```
giadarol, line 814: I think rebinding element_dict breaks the xdeps expressions

xtrack/line.py
```diff
@@ -726,101 +768,177 @@ def get_s_position(self, at_elements=None, mode="upstream"):
...
 
-            if _is_drift(ee):
+            if _is_drift(ee) and not nn in keep:
                 prev_nn = newline.element_names[-1]
                 prev_ee = newline.element_dict[prev_nn]
                 if _is_drift(prev_ee):
                     prev_ee.length += ee.length
                 else:
-                    newline.append_element(ee, nn)
+                    newline.append_element(ee.copy(), nn)
```
giadarol, line 863: Why do you make a copy?

xtrack/line.py
rdemaria: xdeps updates values in `_owner` attribute of the ref. It is possible to replace it to a new dictionary if needed.

freddieknets: I reinitialised the ref; that should do it I believe.

freddieknets:
I made a few fixes:
1. `ee.copy()` is only done in case `inplace=False` (not needed otherwise)
2. var management is correctly initialised / copied in the simplification methods
3. var management is tested for equality in `_lines_equal`
4. the lines used in the simplification tests now have a var management, so that  this is properly tested.

All tests pass on CPU

giadarol:
For some reason it seems to compare to an old version. I'll close and reopen

giadarol:
Followed up in #313

PR #328, opened 2023-03-21
Verbose flag when loading line

freddieknets:
## Description
Very small addition: when loading a line (from a dict, json, or madx sequence), there is now a flag `verbose` (default `True`), and the same for `line.check_aperture()`. 
This is useful when logging the output to a file, as otherwise there will be a few hundred lines with 'loading line xx%'.

## Checklist

Mandatory: 

Optional:


fsoubelet:
As a side note, could we make the loader use a logging function instead of `print`?

giadarol:
> As a side note, could we make the loader use a logging function instead of `print`?

Yes, definitely!

PR #339, opened 2023-04-18
Enable OpenMP support

szymonlopaciuk:
## Description

This PR is part of the effort to enable OpenMP support in Xsuite.

In Xtrack, on the OpenMP context we introduce two loops particle loops:
- an outer loop over batches of particles that is the one being paralellised (this is similar to the GPU behaviour, but for batches instead of single particles)
- an inner loop inside of each element, which loops over individual particles, and which allows for potential gains from vectorisation (this is similar to the serial CPU behaviour)

In addition, at the end of each kernel, particles are reorganised such that they conform to the expectations from the serial CPU behaviour. Nevertheless, this behaviour can be disabled by setting `XT_OMP_SKIP_REORGANIZE` in the config of the line (it should be noted that our current test suite passes with this flag set as well).

Needs xsuite/xobjects#106 and xsuite/xpart#78.

Closes xsuite/xsuite#15.

## Checklist

Mandatory: 

Optional:

### DIFF
--- a/xtrack/base_element.py
+++ b/xtrack/base_element.py
@@ -16,28 +16,31 @@ from .general import _pkg_root
 from .internal_record import RecordIdentifier, RecordIndex, generate_get_record
 
 start_per_part_block = """
-   const int64_t part_id = part0->ipart;  //only_for_context cpu_openmp
-   const int64_t end_id = part0->endpart; //only_for_context cpu_openmp
-   //#pragma omp simd                     //only_for_context cpu_openmp
-   for (int64_t ii=part_id; ii<end_id; ii++){ //only_for_context cpu_openmp
-   
-   int64_t const n_part = LocalParticle_get__num_active_particles(part0); //only_for_context cpu_serial
-   for (int jj=0; jj<n_part; jj+=!!CHUNK_SIZE!!){                 //only_for_context cpu_serial
-    for (int iii=0; iii<!!CHUNK_SIZE!!; iii++){                   //only_for_context cpu_serial
-      int const ii = iii+jj;                                      //only_for_context cpu_serial
-      if (ii<n_part){                                             //only_for_context cpu_serial
+    const int64_t part_id = part0->ipart;  //only_for_context cpu_openmp
+    const int64_t end_id = part0->endpart; //only_for_context cpu_openmp
+    //#pragma omp simd                     //only_for_context cpu_openmp
+    for (int64_t ii=part_id; ii<end_id; ii++) { //only_for_context cpu_openmp
 
-        LocalParticle lpart = *part0;//only_for_context cpu_serial cpu_openmp
-        LocalParticle* part = &lpart;//only_for_context cpu_serial cpu_openmp
-        part->ipart = ii;            //only_for_context cpu_serial cpu_openmp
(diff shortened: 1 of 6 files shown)


xtrack/base_element.py
```diff
@@ -16,25 +16,33 @@
...
-
-        LocalParticle lpart = *part0;//only_for_context cpu_serial cpu_openmp
-        LocalParticle* part = &lpart;//only_for_context cpu_serial cpu_openmp
-        part->ipart = ii;            //only_for_context cpu_serial cpu_openmp
-
-        LocalParticle* part = part0;//only_for_context opencl cuda
+    {
+    const int64_t part_id = part0->ipart;  //only_for_context cpu_openmp
+    const int64_t end_id = part0->endpart; //only_for_context cpu_openmp
+    //#pragma omp simd                     //only_for_context cpu_openmp
```
giadarol, line 22: So we don't need this? If not lets remove this commented line

xtrack/base_element.py
```diff
@@ -16,25 +16,33 @@
...
+    for (int64_t ii=part_id; ii<end_id; ii++) { //only_for_context cpu_openmp
```
giadarol, line 23: Maybe it is more intuitive to rename `part_id` insto `start_id`

xtrack/base_element.py
```diff
@@ -16,25 +16,33 @@
...
+
+    int64_t const n_part = LocalParticle_get__num_active_particles(part0); //only_for_context cpu_serial
+    for (int jj=0; jj<n_part; jj+=!!CHUNK_SIZE!!){                 //only_for_context cpu_serial
+        for (int iii=0; iii<!!CHUNK_SIZE!!; iii++){                //only_for_context cpu_serial
+            int const ii = iii+jj;                                 //only_for_context cpu_serial
```
giadarol, line 28: We could probably remove this nested loops and just loop over `ii`

xtrack/base_element.py
```diff
@@ -16,25 +16,33 @@
...
-      if (ii<n_part){                                             //only_for_context cpu_serial cpu_openmp
-
-        LocalParticle lpart = *part0;//only_for_context cpu_serial cpu_openmp
-        LocalParticle* part = &lpart;//only_for_context cpu_serial cpu_openmp
-        part->ipart = ii;            //only_for_context cpu_serial cpu_openmp
-
-        LocalParticle* part = part0;//only_for_context opencl cuda
+    {
+    const int64_t part_id = part0->ipart;  //only_for_context cpu_openmp
+    const int64_t end_id = part0->endpart; //only_for_context cpu_openmp
```
giadarol, line 21: These names are a bit confusing, as they have nothing to do with the `particle_id`

xtrack/tracker.py
```diff
@@ -940,7 +958,8 @@ def _track_no_collective(
 
         self._check_invalidated()
 
-        if isinstance(self._buffer.context, xo.ContextCpu):
+        if (isinstance(self._buffer.context, xo.ContextCpu)
```
giadarol, line 961: I think this check should stay also the using multiple threads.

giadarol (changes requested):

PR #344, opened 2023-05-04
Add slicing of some thick elements

szymonlopaciuk:
## Description

This allows for enabling the slicing of thick elements in xsuite. Currently bends, quadrupoles, sextupoles, octupoles, and kickers are supported, and two slicing strategies are available: uniform and teapot.

The API is demonstrated in `examples/mad_loader2/job_convert_thick.py`.

## Checklist

Mandatory: 

Optional:


eltos:
I'm curious: Does this work with a line created via `Line.from_sequence`? I.e. can I do something like this 

```python
nodes = [
    Node(3.0, Multipole(length=1, knl=[0, +0.63]), name="quad"),
]
line_thick = Line.from_sequence(nodes, length=10)
line_thin = line_thick.copy()
slicing_strategies = [
    Strategy(slicing=Teapot(1)),
]
line_thin.slice_thick_elements(slicing_strategies)

```

If yes, is the node position assumed to be in the center (such that the element extends from s-length/2 to s+length/2) or beginning (s to s+length)?

giadarol:
Feature introduced in v0.37.0

giadarol:
> 

HI, this applies PR only to "true thick" element types, i.e. those that expose the flag `is thick = True` and that change the s position of the particles. At the moment the elements that do that are Bend, Quadrupole and CombinedFunctionMagnet. Indeed, we did not think of testing their behavior within `Sequence`. Would you be willing to have a look at that?

To avoid forgetting this point, I add an item to this issue https://github.com/xsuite/xsuite/issues/351

eltos:
Ah, I see, thank you for the explanation.

In this case it will not work, see [here](https://github.com/xsuite/xtrack/blob/6ee038bc81236d22e017e0ce689f039a50abc830/xtrack/line.py#L320-L322)
```
            if _is_thick(node.what): 
                 raise NotImplementedError( 
                     f'Thick elements currently not implemented: {node.name}')
```

PR #349, opened 2023-06-09
v0.36.3

giadarol:
**Changes:**
 - More accurate calculation of dispersion and crab dispersion
 - In twiss `method="4d"` freezes the energy variables (but not zeta).
 - Fixes in management of cached and pre-built kernels.
 - Fixes in twiss orientation management


giadarol:
Will be handled in v0.36.4 together with other modifications

PR #370, opened 2023-07-20
Element-by-element monitor in collective mode

anabramo:
## Description

Made changes to tracker.py to allow using ONE_TURN_EBE flag is collective trackers. This is done by passing a monitor instance to the non-collective tracker parts in the collective tracker. In the case of a regular TBT monitor, the monitor instance is not passed to the tracker parts.

## Checklist

Mandatory: 

Optional:

- [x ] The code I wrote follows good style practices (see [PEP 8](https://peps.python.org/pep-0008/) and [PEP 20](https://peps.python.org/pep-0020/)).


giadarol:
That's great, thanks a lot!

tests/test_monitor.py
```diff
@@ -167,3 +167,58 @@ def test_last_turns_monitor(test_context):
     assert np.all(monitor.x == np.array([[0,0,2,1,0],[3,5,7,9,11],[0,-2,-4,-6,-8],[20,23,26,29,32]]))
 
 
+@for_all_test_contexts
+def test_collective_monitor(test_context):
```
giadarol, line 171: To be renamed `test_collective_ebe_monitor`

giadarol (approved):

tests/test_monitor.py
anabramo: I've renamed the test as per the suggestion and added a check for None in the monitor definition.

PR #373, opened 2023-08-01
Add metadata to multiline

ColasDroin:
Added a metadata attribute to the Multiline class. This attribute is a dictionary containing the collider configuration, e.g. as in [here](https://github.com/xsuite/example_DA_study/blob/main/master_study/master_jobs/2_configure_and_track/config.yaml). This should be helpful for reproducibility of simulations, allowing to embed the configuration of a collider in the corresponding json file.

A unit test has been implemented, in `test_multiline.py`. Of note, the previous unit test for the Multiline class was highly minimal and merged with the orbit correction test. Consequently, I've split `test_multiline_and_orbit_correction.py` into `test_multiline.py` and `test_orbit_correction.py`. All tests are passing.

### DIFF
--- a/tests/test_line.py
+++ b/tests/test_line.py
@@ -541,6 +541,8 @@ def test_to_dict():
 
     assert result['elements']['d']['__class__'] == 'Drift'
     assert result['elements']['d']['length'] == 1
+    
+    assert result['metadata'] == line.metadata
 
 
 def test_from_dict_legacy():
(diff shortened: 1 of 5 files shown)


ColasDroin:
Updated PR after discussion with @giadarol such that the metadata is now always an attribute of Multiline and Line, as an empty dic by default. Added corresponding unit tests, which are passing.

## Notes

- I did not explicitly make the metadata a property of Multiline or Line, not to complicate the code while I believe we don't need a getter/setter for now. I could change this by adding a ```_metadata``` attribute that would be accessed through a ```metadata``` property.
- As a dictionary object, the metadata tends to be passed as a reference when using the ```from_dict()```, ```to_dict()``` methods. This means that copying a collider does not explicitly copy the metadata, only the reference, such that modifying the metadata in one collider will also modify the metadata in the copy. This could easily be solved by doing a ```copy.deepcopy()```of the dictionnary in the ```to_dict()```method, but I'm not sure if we want to involve the ```copy```package. Tell me if I should update this.

giadarol:
I think it's a good idea to use deepcopy. For me you can go ahead and do it that way

ColasDroin:
Now implemented deepcopy() in to_dict() methods, all relevant tests are passing.

PR #374, opened 2023-08-03
Added monitor as attribute of Footprint when computing footprint

ColasDroin:
Very minimal change to add monitor as an attribute of the Footprint class, when computing the footprint. This is needed to compute/compare the footprints with alternative libraries such as pyNAFF or NAFFlib.


ColasDroin:
Updated the code following discussion with @giadarol , adding keep_tracking_data as an argument of the Footprint class to avoid always saving the (heavy) monitor object.

ColasDroin:
Updated code as I realized I didn't update unit test to accommodate for the footprint flag. Fixed the implementation accordingly. Test is passing.

PR #380, opened 2023-08-11
Implement a solenoid

szymonlopaciuk:
## Description

Implement a thick solenoid in Xsuite.

## Checklist

Mandatory: 

Optional:

### DIFF
--- a/tests/test_elements_thick.py
+++ b/tests/test_elements_thick.py
@@ -743,11 +743,6 @@ def test_import_thick_with_apertures_and_slice():
         (-0.1, 0, 0.9),
         (0, 0, 0.9),
         (0.13, 0, 1.6),
-        # thin:
-        pytest.param(-0.1, 0.12, 0, marks=pytest.mark.xfail(reason='MAD-X bug, see #1200')),
-        pytest.param(0, 0.12, 0, marks=pytest.mark.xfail(reason='MAD-X bug, see #1200')),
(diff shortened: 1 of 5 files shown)


rdemaria:
I recommend performing comparisons with small relativistic beta. A benchmark against mad-x is useful but I would not use it as a test.

You could add a test checking that a thick solenoid with zero field gives the same results of a drift for px!=py!=0.

Do you have a  write out of the equation you implemented? Do they agree with what is in paragraph 12.5.2 in https://inspirehep.net/literature/487284 ?

szymonlopaciuk:
I can put in some tests for a small beta, I can also check if the solenoid behaves like a drift with zero field, although having it as a test is maybe not very strong, as the code explicitly catches that case and does a drift.

I have looked at the book, but in the end the code that is here (and in MAD-X) is not obviously the same as there. I could implement those equations and check for agreement there.

rdemaria:
The check with zero fields can catch bugs. I expect the equations from the book and MAD-X (https://github.com/MethodicalAcceleratorDesign/MAD-X/pull/1129)  to implement the same solution as it is exact. Do you suspect they can be different?

rdemaria:
For the thin solenoid, you can have a look at:

https://cds.cern.ch/record/281283/files/sl-95-012.pdf
eq. 3.37

which obtains a formulation under some approximations

PR #402, opened 2023-10-17
New element Henon map

deveres99:
## Description
A new beam element that simulates a Henon map is added. The Henon map can be created with arbitrary multipoles, but all assumed to be at the same phase advance. The map also correctly handles off-momentum particles, chromaticity and dispersion. The element is useful when theoretical studies on a relatively simple mathematical model are needed (e.g. for DA studies).

Closes # .

## Checklist

Mandatory: 

Optional:


xtrack/beam_elements/elements.py
```diff
@@ -84,6 +84,171 @@ def add_slice(weight, container, thick_name, slice_name, _buffer=None):
...
+    norm: int
+        1 if input coordinates are already normalised, 0 if not. Default is ``0``.
+
+    Comments
+    --------
+
+    - The user provides omega_x and omega_y, their sin and cos is calculated at 
+      initialisation.
+    - The user provides "multipole_coeffs", an array of floats, that contains the 
+      strength of the multipoles present in the map. Default is ``[0]``.
```
szymonlopaciuk, line 155: I would put these in the parameters section. More generally, the parameters section should contain the input allowed from the user.

On a related note, you accept both `omega_x` and `sin_omega_x` (and analogously for cos and y). It could be worth it to add a consistency check, or only accept angle if and only if no trig values are given (the latter may work, since you don't store the angle, otherwise it would be a problem when loading the values back from dict; for the former you can look at https://github.com/xsuite/xtrack/blob/main/xtrack/beam_elements/elements.py#L424 and use the helper function.) Likely the same comment applies for `multipole_coeffs`.

xtrack/beam_elements/elements.py
```diff
@@ -84,6 +84,171 @@ def add_slice(weight, container, thick_name, slice_name, _buffer=None):
...
+
+    isthick = False
+    behaves_like_drift = False
+
+    _extra_c_sources = [_pkg_root.joinpath('beam_elements/elements_src/henonmap.h')]
+
+    def __init__(self, omega_x = 0.,
+                       omega_y = 0.,
+                       n_turns = 1, 
+                       twiss_params = [0., 1., 0., 1.],
```
szymonlopaciuk, line 189: Using lists as default arguments in Python should be avoided (https://docs.python-guide.org/writing/gotchas/#mutable-default-arguments), as lists are mutable.

This won't cause problems in your particular case, but for good practice consider using a tuple instead, or `None` and a check for `is None` in the body of the function.

xtrack/beam_elements/elements_src/henonmap.h
```diff
@@ -0,0 +1,184 @@
...
+                double sin_domega_y = sin(domegay * delta);
+                curr_cos_omega_y = cos_omega_y * cos_domega_y - sin_omega_y * sin_domega_y;
+                curr_sin_omega_y = sin_omega_y * cos_domega_y + cos_omega_y * sin_domega_y;
+            }
+
+            double x_hat_new, px_hat_new, y_hat_new, py_hat_new;
+            x_hat_new = curr_cos_omega_x * (x_hat - x_hat_f) + curr_sin_omega_x * (px_hat - px_hat_f + fx) + x_hat_f;
+            px_hat_new = -curr_sin_omega_x * (x_hat - x_hat_f) + curr_cos_omega_x * (px_hat - px_hat_f + fx) + px_hat_f;
+            y_hat_new = curr_cos_omega_y * y_hat + curr_sin_omega_y * (py_hat + fy);
+            py_hat_new = -curr_sin_omega_y * y_hat + curr_cos_omega_y * (py_hat + fy);
```
szymonlopaciuk, line 151: These could be `const`, but maybe they aren't needed? You could write directly to the target variables and put a comment like "update normalised coordinates".

tests/test_henonmap.py
```diff
@@ -0,0 +1,88 @@
...
+                             norm = False)
+    line_test = xt.Line(elements=[henon_test], element_names=["henon"])
+    line_test.build_tracker(_context=test_context)
+
+    line_test.track(p_test)
+
+    x_out = test_context.nparray_from_context_array(p_test.x)[np.argsort(test_context.nparray_from_context_array(p_test.particle_id))]
+    px_out = test_context.nparray_from_context_array(p_test.px)[np.argsort(test_context.nparray_from_context_array(p_test.particle_id))]
+    y_out = test_context.nparray_from_context_array(p_test.y)[np.argsort(test_context.nparray_from_context_array(p_test.particle_id))]
+    py_out = test_context.nparray_from_context_array(p_test.py)[np.argsort(test_context.nparray_from_context_array(p_test.particle_id))]
```
szymonlopaciuk, line 83: You can use `p_test.sort(interleave_lost_particles=True)` above and then there should be no need for argsort.

szymonlopaciuk (changes requested):
I discussed with @giadarol and it would be really good to have an entry in the physics guide, also for the review.

The code itself is very nice, I wrote down a few small things I noticed for your consideration.

tests/test_henonmap.py
deveres99: I tried to replace this with the following:
```
    p_test.sort(interleave_lost_particles=True)
    x_out = test_context.nparray_from_context_array(p_test.x)
    px_out = test_context.nparray_from_context_array(p_test.px)
    y_out = test_context.nparray_from_context_array(p_test.y)
    py_out = test_context.nparray_from_context_array(p_test.py)
```
but this way the GPU test fails with
```
        if not isinstance(self._buffer.context, xo.ContextCpu):
>           raise NotImplementedError('Sorting only works on CPU for now')
E           NotImplementedError: Sorting only works on CPU for now

../../xpart/xpart/particles/particles_base.py:711: NotImplementedError
```
Is there another smarter way to do the sorting such that it works on GPU as well, or can the original version of the test stay?

deveres99: I tried to replace this with the following:
```
    p_test.sort(interleave_lost_particles=True)
    x_out = test_context.nparray_from_context_array(p_test.x)
    px_out = test_context.nparray_from_context_array(p_test.px)
    y_out = test_context.nparray_from_context_array(p_test.y)
    py_out = test_context.nparray_from_context_array(p_test.py)
```
but this way the GPU test fails with
```
        if not isinstance(self._buffer.context, xo.ContextCpu):
>           raise NotImplementedError('Sorting only works on CPU for now')
E           NotImplementedError: Sorting only works on CPU for now

../../xpart/xpart/particles/particles_base.py:711: NotImplementedError
```
Is there another smarter way to do the sorting such that it works on GPU as well, or can the original version of the test stay?

xtrack/beam_elements/elements.py
deveres99: Sorry, I thought the parameters should be the _xofields attributes, so I updated the comment to contain under "Parameters" the user input. However, I did write the element such that it runs as fast as possible, i.e. avoiding any unnecessary trig evaluations etc. For this reason I don't store the `omega_x`, only its sine and cosine. Thus I also don't accept e.g. `sin_omega_x` as input. The user can still change the tune during a many-turn simulation for example, but they must then provide `sin_omega_x` and `cos_omega_x` directly (i.e. `henon.sin_omega_x = my_new_value`). Likewise, they can change multipole strength, but they must provide the `fx_x_coeffs`, `fy_x_coeffs` etc. instead of simply the integrated multipole strength they provided at initialization. This is not a super likely usecase, but I myself do this sometimes, so I thought it important to point this out, and I added a "Comment" section with a detailed explanation of the `_xofields` attributes and a "Notes" section that should make it clear for the user how they can change these properties of the map if they need to.

xtrack/beam_elements/elements.py
deveres99: I changed the default to None and throw a value error if nothing is provided.

xtrack/beam_elements/elements_src/henonmap.h
deveres99: I made them `const` now, but they are needed, because e.g. I use the value of `x_hat` to calculate `px_hat_new` etc.

xtrack/beam_elements/elements_src/henonmap.h
```diff
@@ -0,0 +1,184 @@
...
+// This file is part of the Xtrack Package.  //
+// Copyright (c) CERN, 2021.                 //
+// ######################################### //
+
+#ifndef XTRACK_HENONMAP_H
+#define XTRACK_HENONMAP_H
+
+
+/*gpufun*/
+void Henonmap_track_local_particle(HenonmapData el, LocalParticle* part0){
```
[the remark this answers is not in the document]
deveres99: I have opened a pull request for the physics manual update, which now contains the Henon map maths: https://github.com/xsuite/xsuite/pull/415

PR #409, opened 2023-10-31
Missing backtrack

freddieknets:
## Description

Very small changes:
- `SimpleThinQuadrupole` and `SimpleThinBend` were missing a backtrack flag
- in `twiss`, if particles are lost also their state and the element where they are lost are outputted in the error for easier debugging

## Checklist

Mandatory: 

Optional:

- [X] The code I wrote follows good style practices (see [PEP 8](https://peps.python.org/pep-0008/) and [PEP 20,](https://peps.python.org/pep-0020/)).

### DIFF
--- a/xtrack/twiss.py
+++ b/xtrack/twiss.py
@@ -754,7 +754,8 @@ def _twiss_open(line, twiss_init,
     if not _continue_if_lost:
         assert np.all(ctx2np(part_for_twiss.state) == 1), (
             'Some test particles were lost during twiss! '
-          + f'(state: {np.unique(ctx2np(part_for_twiss.state))})')
+          + f'(state {np.unique(ctx2np(part_for_twiss.state))}, '
+          + f'at element {np.unique(ctx2np(part_for_twiss.at_element))})')
 
     if twiss_orientation == 'forward':
         i_start = ele_start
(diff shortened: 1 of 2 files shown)


xtrack/twiss.py
```diff
@@ -770,7 +772,9 @@ def _twiss_open(line, twiss_init,
     recorded_state = line.record_last_track.state[:, i_start:i_stop+1].copy()
     if not _continue_if_lost:
         assert np.all(recorded_state == 1), (
-            'Some test particles were lost during twiss!')
+             'Some test particles were lost during twiss! '
+          + f'(state {np.unique(recorded_state)}, '
+          + f'at element {np.unique(line.record_last_track.at_element[:, i_start:i_stop+1].copy())})')
```
giadarol, line 777: Doesn' this display the element index for all particles, including those that are not lost?

giadarol (changes requested):

xtrack/twiss.py
freddieknets: Indeed, but as it’s wrapped in `np.unique` it is literally only one extra outputted list item. I chose to do so as I didn’t want to slow down the calculation just for error output. But if you want, I can add it.

giadarol: OK with me

giadarol (approved):

PR #413, opened 2023-11-07
Feature/line seg map multiplebucket

giadarol:


xtrack/beam_elements/elements.py
```diff
@@ -1794,6 +1794,7 @@ class LineSegmentMap(BeamElement):
         'longitudinal_mode_flag': xo.Int64,
         'qs': xo.Float64,
         'bets': xo.Float64,
+        'bucket_length': xo.Float64,
```
giadarol, line 1797: @xbuffat Is it needed to introduce a new input? Can't we use `frequency_rf`?

xtrack/beam_elements/elements.py
```diff
@@ -1987,9 +1993,12 @@ def __init__(self, length=None, qx=0, qy=0,
             assert voltage_rf is None
             assert frequency_rf is None
             assert lag_rf is None
+            if bucket_length == None:
```
giadarol, line 1996: (if we keep bucket length)
We need to check that, if it is provided, we are in the right longitudinal mode. Otherwise we need to raise an exception

xbuffat: Done, the bucket_length is kept only for the mode 'fixed_qs', an error is raised otherise.

giadarol:
It would prefer not to rely on ducktrack only for the tests. We could make a consistency test where we use the code of Lorenzo to generate a few bunches, measure bunch length and z mean with the xfields, track a few synchrotron period with this map and check that the bunch does not move.

PR #425, opened 2023-12-08
RF-Track interface

alatina:
## Description

This PR implements the interface between Xsuite and RF-Track. One example showing two use cases has been added, as well as a test unit. A short description has been added to the new element.

## Checklist

Mandatory: 

Optional:


giadarol:
The implementation and the tests are very clear. Thank you

It would be good to improve  the test coverage, by adding:
- [ ] A case with a non-uniform field (e.g. a quadrupole).
- [ ] A test where lost particles from Xsuite are passed to RFTrack, asserting that they are neglected by RFTrack intensity effects and are left untouched when back to Xsuite.
- [ ] A test where particles are lost on the RFtrack side, checking that they are properly handled by Xsuite.
- [ ] A test with a time-dependent field (e.g. an RF cavity) to check that the time conversion is correct (I would use a case with beta0 << 1 to make sure that the code behaves also in non-ultrarelativistic cases).
 - [ ] A test where the energy of the particles changes in RFTrack, to make sure that energy conversions are handled properly.

alatina:
I added three examples:
- A case of non-uniform field
- A case of a 3D solenoid with time-dependent tracking
- A case where lost particles are passed back and forth between the codes, and new losses occur in RF-Track

alatina:
I just added an example of tracking through the RF field map of a travelling wave structure. 

This led to fixing a bug: the update of the reference momentum was missing.

I think now the PR can be accepted!

PR #429, opened 2023-12-18
Add pandas dependency

fsoubelet:
## Description

Closes https://github.com/xsuite/xsuite/issues/444.

This is a simple addition of `pandas` as a dependency of the package. I have checked around and the only use that is made is to call `pandas.DataFrame`, with no special function called from it so we should be relatively fine regarding the minimum version.

I have set `>=2.0` so we should be good for a long while before another major release comes out.

## Checklist

Mandatory: 

Optional:

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -41,6 +41,7 @@ setup(
     include_package_data=True,
     install_requires=[
         'numpy>=1.0',
+        "pandas>=2.0",
         'scipy',
         'xobjects',
         'xpart',


fsoubelet:
@giadarol this should be an easy merge ;)

giadarol:
> @giadarol this should be an easy merge ;)

Already in for the next release

giadarol (approved):

PR #445, opened 2024-02-13
Cut the line at specified s coordinates

szymonlopaciuk:
## Description

Given a list of $s$ coordinates, make cuts in the line at those positions. This uses the slicing module as to not repeat the logic used to preserve compounds. A new slicing scheme called `Custom` is introduces, which allows to cut elements at $s$ positions relative to the start of the element.

## Checklist

Mandatory: 

Optional:

### DIFF
--- a/examples/lhc_thick/004_slice_every_x_meters.py
+++ b/examples/lhc_thick/004_slice_every_x_meters.py
@@ -1,0 +1,29 @@
+import matplotlib.pyplot as plt
+from pathlib import Path
+from time import time
+
+import xtrack as xt
+import numpy as np
+
+all_num_cuts = np.logspace(np.log2(10), np.log2(50_000), base=2, num=15)
+time_to_cut = []
+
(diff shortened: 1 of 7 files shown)


szymonlopaciuk:
I have analysed the performance of this solution on the LHC example, and the results can be seen in the graphs below. It is not what I expected: considering that finding a slicing strategy for an element is in O(num_strategies) = O(min(num_cuts, num_cut_elements)) $\subseteq$ O(min(num_cuts, num_elements)), I expected the shape of the graph to look more like O(min(num_cuts, num_elements) * num_elements), i.e. more like a parabola for small x = num_cuts. While I'm not sure yet where this shape for small x comes, it seems that for large x we do tend to be linear, which is not so bad.

Nevertheless, there should be some possibility of quite some improvement in the strategy matching, which I will try next, and implement if the results look promising. 

Figure_1
Figure_3
Figure_4

PR #453, opened 2024-03-01
Fix prebuilt kernels

freddieknets:
## Description

The current implementation of prebuilding kernels for (non-tracker) kernels and the attaching of non-particle kernels to a `BeamElement` do not work (the former will not use prebuilt kernels, and the latter will even raise an error when trying to instantiate such an object). This fix addresses that.

Changes:
- The variable `prebuilt_kernels_path` was not passed correctly to `BeamElement.compile_kernels`, making it fail (as it was then passed to `xo.Struct` which doesn't know the variable).
- The `PerParticlePyMethod` did not have this flag at all, making it skip prebuilt kernels
- When looking for a suitable kernel, an empty config would not find anything, as the prebuilt kernels are always stored with a specific config. Now it defaults to the default config of a `Line`.
- Added `Particles._init_random_number_generator` to prebuilt kernels.
- Added a flag `needs_rng` to a `BeamElement`, such that element classes can impose the need for a random generator (like is the case for an `EverestCollimator`). This removes the need for the user to call `Particles._init_random_number_generator()` which feels a bit hacky.

The latest version of `Xcoll` (to be released after this PR is merged)  has a few elements that rely on these changes, hence the `Xcoll` tests act implicitly as tests for this behaviour.

## Checklist

Mandatory: 

Optional:

### DIFF
--- a/xtrack/base_element.py
+++ b/xtrack/base_element.py
@@ -290,28 +290,25 @@ class BeamElement(xo.HybridClass, metaclass=MetaBeamElement):
                     local_particle_src=Particles.gen_local_particle_api()))
         context = self._context
         cls = type(self)
+        prebuilt_kernels_path = kwargs.pop('prebuilt_kernels_path')
         if context.allow_prebuilt_kernels:
-            import xtrack as xt
-            from xtrack.prebuild_kernels import (
-                get_suitable_kernel,
-                XT_PREBUILT_KERNELS_LOCATION
-            )
+            from xtrack.prebuild_kernels import get_suitable_kernel
             # Default config is empty (all flags default to not defined, which
             # enables most behaviours). In the future this has to be looked at
             # whenever a new flag is needed.
             _default_config = {}
             _print_state = Print.suppress
-            # Print.suppress = True
+            Print.suppress = True
             classes = (cls._XoStruct,) + tuple(extra_classes)
             kernel_info = get_suitable_kernel(
                 _default_config, classes
             )
-            # Print.suppress = _print_state
+            Print.suppress = _print_state
             if kernel_info:
                 module_name, _ = kernel_info
                 kernels = context.kernels_from_file(
                     module_name=module_name,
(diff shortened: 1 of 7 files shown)


szymonlopaciuk:
We should catch up on Monday but I just wanted to comment on one of the points.

The problem with assuming a default config of the line for the elements, is that it's not the right config for all elements. When I assumed that, tests for xfields would fail, as e.g. `NO_BHABHA...` set to true is not the correct setting -- but an empty config is. We've discussed with Gianni, and the reason why config flags often start with `NO_` is that their absence is assumed for the full behaviour of an element.

Nevertheless you're completely right that the current implementation doesn't work. I'm not sure what happened in my testing that it "picked up" the kernels, but reading the code now it seems it shouldn't... Maybe a subset check for the config compatibility would be the right thing to do, or assuming that the missing keys must default to True?

freddieknets:
You are very right, I had overseen that tracking on the element of course does not set any compiler flags. This can easily be addressed by providing kernels for empty configs as well. Testing it now.

freddieknets:
All tests pass for all packages

xtrack/base_element.py
```diff
@@ -273,6 +273,8 @@ class BeamElement(xo.HybridClass, metaclass=MetaBeamElement):
     allow_backtrack = False
     skip_in_loss_location_refinement = False
     prebuilt_kernels_path = XT_PREBUILT_KERNELS_LOCATION
```
szymonlopaciuk, line 275: I know that this is something that I wrote, but do we actually need this? I don't think so, but I wanted to check with you @freddieknets, I don't want to break something again. It seems that in a roundabout way it's used by the `PyMethod`s which then pass it to `compile_kernels` (could be used directly...), but in any case it's never anything but the default? Right?

Anyway, I added a little test, and all of them are passing.

szymonlopaciuk (approved):

PR #457, opened 2024-03-09
Smarter OpenMP

freddieknets:
## Description

For simulations where particles can die early on (and where this happens often), like in loss map simulations, the current Open MP implementation is not optimal, as in some threads more particles will have died than others, and the freed resources are not re-assigned (as we are in a big loop over threads, outside of the loop over turns).

It is probably possible to rewrite the logic completely in order to address this smartly, with the particles in a stack where each thread tracks a single particle for all turns (or only over the non-collective parts in case of a collective tracker), and when finished or dead, the thread takes the next particle from the stack to be tracked. This would also be a smart implementation for tracking with GPUs in this scenario.

But for now, there is a very simple fix that is almost as good:
- track turn-by-turn (or track with `with_progress=1`)
- at the start of tracking, assign the OpenMP threads based on the number of alive particles instead of the total capacity.

This PR implements the latter with, as said, very minimal changes. But the total gain in a standard loss map simulation is almost a factor 4 in tracking (29s instead of 110s for 50.000 particles and 200 turns), which absolutely makes it worth it (especially for such a small change).

I'm running the tests now, and so far they all passed.

EDIT: all tests pass

## Checklist

Mandatory: 

Optional:

### DIFF
--- a/xtrack/base_element.py
+++ b/xtrack/base_element.py
@@ -105,16 +105,17 @@ def _generate_per_particle_kernel_from_local_particle_function(
 '''
                              int64_t flag_increment_at_element,
                 /*gpuglmem*/ int8_t* io_buffer){
-            const int num_threads = omp_get_max_threads();                                 //only_for_context cpu_openmp
-            const int64_t capacity = ParticlesData_get__capacity(particles);               //only_for_context cpu_openmp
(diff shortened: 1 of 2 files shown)


giadarol (approved):
this was in fact a bug. your implemented solution was the intended implementation

PR #469, opened 2024-04-02
Fix loading compounds from Madx for repeated element names

szymonlopaciuk:
## Description

When loading a lattice from Madx where names of elements repeat, only the first compound is loaded properly. This fixes the issue and loads the compounds properly.

## Checklist

Mandatory: 

Optional:


giadarol:
Obsolete

PR #484, opened 2024-04-23
Io buffer and loss rename

freddieknets:
## Description

Three quick but important changes:
- If no `io_buffer` is specified at the tracker initialisation, but (some of) the line elements have a `io_buffer` specidied, this one is automatically assigned to the tracker (it is checked that they all point to the same `io_buffer`)
- Renamed `allow_backtrack` into `allow_loss_refinement`
- Drift slices were missing `allow_loss_refinement=True`

## Checklist

Mandatory: 

Optional:


giadarol:
already merged

PR #499, opened 2024-05-28
In test_twiss_range_start_end use a tight but inexact assert; use xo.assert_... everywhere

szymonlopaciuk:
## Description

Change the assertion is `test_twiss.py::test_twiss_range_start_end` to use an inexact comparison, as currently the tests are failing with very small differences.

Use `xo.assert_allclose` whenever possible in the tests: I had to relax two `rtols` to make tests pass compared to before:
- in `test_vs_madx.py::test_twiss_and_survey`
- in `test_xmask_orbit_correction.py::test_orbit_correction`

## Checklist

Mandatory: 

Optional:

- x ] The code I wrote follows good style practices (see [PEP 8](https://peps.python.org/pep-0008/) and [PEP 20](https://peps.python.org/pep-0020/)).

### DIFF
--- a/tests/test_xmask_orbit_correction.py
+++ b/tests/test_xmask_orbit_correction.py
@@ -1,9 +1,10 @@
 import json
 import pathlib
 
-import numpy as np
-import xtrack as xt
 from cpymad.madx import Madx
+
+import xobjects as xo
+import xtrack as xt
 from xobjects.test_helpers import for_all_test_contexts
 
 test_data_folder = pathlib.Path(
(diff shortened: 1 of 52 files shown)


tests/test_xmask_orbit_correction.py
```diff
@@ -228,21 +229,21 @@ def test_orbit_correction(test_context):
 
     tw = collider.lhcb1.twiss()
 
-    assert np.isclose(tw['px', 'ip1'], 250e-6, atol=1e-8)
```
szymonlopaciuk, line 231: rtol=0, atol=5e-8

szymonlopaciuk: Maybe make atol the first argument of the two and set rtol=0 by default.

PR #500, opened 2024-05-30
Add multipolar components to solenoid

szymonlopaciuk:
## Description

- Add `knl`, `ksl` and related fields to the solenoid element to implement multipolar components.
- Also a small amount of refactoring in `xtrack/beam_elements/slice_elements.py`.
- Also see a comment below.

## Checklist

Mandatory: 

Optional:

### DIFF
--- a/xtrack/beam_elements/elements_src/solenoid.h
+++ b/xtrack/beam_elements/elements_src/solenoid.h
@@ -13,12 +13,14 @@ void Solenoid_track_local_particle(SolenoidData el, LocalParticle* part0) {
     double length = SolenoidData_get_length(el);
     double ks = SolenoidData_get_ks(el);
     int64_t radiation_flag = SolenoidData_get_radiation_flag(el);
+    double factor_knl_ksl = 1;
 
     #ifdef XSUITE_BACKTRACK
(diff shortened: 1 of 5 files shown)


xtrack/beam_elements/elements_src/solenoid.h
```diff
@@ -13,12 +13,14 @@ void Solenoid_track_local_particle(SolenoidData el, LocalParticle* part0) {
...
     double ks = SolenoidData_get_ks(el);
     int64_t radiation_flag = SolenoidData_get_radiation_flag(el);
+    double factor_knl_ksl = 1;
 
     #ifdef XSUITE_BACKTRACK
         length = -length;
+        factor_knl_ksl = -1;
     #endif
 
-    #ifndef XTRACK_SOLENOID_NO_SYNRAD
```
szymonlopaciuk, line 21: Also: does this do the right thing? If `XTRACK_SOLENOID_NO_SYNRAD` is defined then the variables are not defined and the compilation will fail. Also these values are passed as pointers to `Solenoid_thick_track_single_particle` but their updated values were not used (now they are though -- is that a correct way?).

PR #515, opened 2024-08-12
Experimental: First Line Parser

szymonlopaciuk:
## Description

Build a first iteration of a fast parser for a MAD-X *inspired* sequence language, currently capable of storing and loading Lines and Multilines. The expressions are kept in place, close to their corresponding elements, as opposed to the way data is structured in json. As an example, here's a "FODO":

```
cell_l = 1;
knl_f = 1;
knl_d = -1;
angle = 15 * raddeg;
h = angle / cell_l;
k0 = h;

line1: beamline;
    particle_ref, p0c = 7e9, q0 = 1, mass0 = pmass;
    b1: Bend, k0 = k0, h = h, length = cell_l;
    qf1: Multipole, knl = {0, knl_f};
    d12u: Drift, length = cell_l / 2;
    m12: Marker;
    d12d: Drift, length = cell_l / 2;
    qd2: Multipole, knl = {0, knl_d};
endbeamline;
```

Twiss default, configs, and metadata can be included. The parser is build using Bison+Flex, which compiles the grammar into C files, which interface with the rest of the parser written in Cython. On a standard HL-LHC lattice the read speed is <1.5x the json loading speed, which is somehow a necessary slow down due to the fact that we need to convert the different data layout -- still further speedup opportunities exist.

This has only been tested with Xtrack so far, and an environment flag `XTRACK_TEST_XLD=1` can be switched on in order to make a JSON->XLD->JSON round-trip every time a json file is loaded. When run as part of the test suite this can be the first step to ensure we do not break things in the future, even though currently this is experimental.

## Checklist

Mandatory: 

- [x] **ONLY XTRACK** tests have been checked for now

Optional:


giadarol:
Decided to stick to python

PR #536, opened 2024-10-03
Added xcoll linetools and element classes to Line.

freddieknets:
## Description

Verify simple PR to expose some of the Xcoll API to the `Line`.

## Checklist

Mandatory: 

Optional:


giadarol (changes requested):
create in Xcoll

```
class LineXcoll:

    def _init__(self, line):
        self.line=line

    def ....
```

xtrack/line.py
```diff
@@ -886,6 +886,19 @@ def particle_ref(self, particle_ref):
             self.particle_ref.t_sim = (
                 self.get_length() / self.particle_ref._xobject.beta0[0] / clight)
 
+    @property
+    def scattering(self):
```
giadarol, line 890:
```python 
    @property
     def scattering(self):
         if not hasattr(self, '_scattering') or self._scattering is None:
             try:
                 from xcoll.line_tools import XcollLineAPI
                 self._scattering = XcollLineAPI(line=self)
             except ImportError as error:
                 raise ImportError("Please install Xcoll to use this feature.") from error

         return self._scattering

```

giadarol (changes requested):

PR #545, opened 2024-10-17
typo at_element -> at_turn in get_normalized_coordinates and reshape …

pbelange:
typo in `twiss_init.get_normalized_coordinates`. Also when `XX_norm.ndim == 3`, the return table was not working because all the arrays were 2D.  `flatten` fixes it. (I think this ONLY happens if the passed`particles` is in fact a monitor object, with turn-by-turn data)

### DIFF
--- a/xtrack/twiss.py
+++ b/xtrack/twiss.py
@@ -2816,7 +2816,7 @@ class TwissInit:
 
         part_id = ctx2np(particles.particle_id).copy()
         at_element = ctx2np(particles.at_element).copy()
-        at_turn = ctx2np(particles.at_element).copy()
+        at_turn = ctx2np(particles.at_turn).copy()
         x_norm = ctx2np(particles.x).copy()
         px_norm = x_norm.copy()
         y_norm = x_norm.copy()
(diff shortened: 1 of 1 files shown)


pbelange:
Also side note: I personally like to pass a `monitor` and get the normalized coordinates seamlessly... to do so the `table` which is returned is not ideal because it has a different shape than the `monitor` which is passed to the function. My personal solution is to do something like this (suggestion):
```python

# Within a custom class which has the twiss_init and the emittances stored:
#...
    def phys2norm(self,monitor):

        table = self.twiss_init.get_normalized_coordinates(monitor,nemitt_x=self.nemitt_x, nemitt_y=self.nemitt_y,nemitt_zeta=self.nemitt_zeta)

        monitor_norm = monitor.copy()
        norm_vars = tuple((tt,nn+'_norm') for tt,nn in xt.Particles.per_particle_vars+((xo.Float64, 'pzeta'),) if nn in ['x','y','zeta','px','py','pzeta'])
        for tt, nn in norm_vars:
            setattr(monitor_norm, nn, xt.monitors.particles_monitor._FieldOfMonitor(name=nn))

        monitor_norm.x_norm = table.x_norm.reshape(monitor.x.shape)
        monitor_norm.y_norm = table.y_norm.reshape(monitor.y.shape)
        monitor_norm.zeta_norm = table.zeta_norm.reshape(monitor.zeta.shape)
        monitor_norm.px_norm = table.px_norm.reshape(monitor.px.shape)
        monitor_norm.py_norm = table.py_norm.reshape(monitor.py.shape)
        monitor_norm.pzeta_norm = table.pzeta_norm.reshape(monitor.pzeta.shape)
        
        return monitor_norm
```

PR #562, opened 2024-12-03
Add time-dependent noise handling inside C implementation of Elens beam element

carlidel:
## Description

As HEL require fast testing with different noise patterns, it is of interest to implement directly in the C code the possibility of handling a time-dependent noise array, representing different realization/operation patterns of HEL.

The code simply implements a float array in the xobjects buffer that the C code then reads depending on the particle at_turn value. Different modes for handling out-of-lenght indexes have been implemented.

A consistency check has been performed. If you think it is of interest, I can include it in the test files.

I have updated the HEL docstring accordingly.  

## Checklist

Mandatory: 

- [ X ] I have added tests to cover my changes
- [ X ] All the tests are passing, including my new ones
- [ X ] I described my changes in this PR description

Optional:

- [ X ] The code I wrote follows good style practices (see [PEP 8](https://peps.python.org/pep-0008/) and [PEP 20](https://peps.python.org/pep-0020/)).
- [ X ] I have updated the docs in relation to my changes, if applicable
- [ X ] I have tested also GPU contexts

### DIFF
--- a/xtrack/beam_elements/elements.py
+++ b/xtrack/beam_elements/elements.py
@@ -172,7 +172,19 @@ class Elens(BeamElement):
         Array of coefficients of the polynomial. Default is ``[0]``.
     polynomial_order : int
         Order of the polynomial. Default is ``0``.
-
+    noise: array
+        Array of noise values. Default is ``[1.0]``. Use this to model different
+        on/off states of the elens for each turn. The length of the noise array
+        should be equal to the number of turns you want to simulate. Otherwise,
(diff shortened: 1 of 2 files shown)


giadarol:
Thank you! Would be nice to have a test added....

carlidel:
Thanks! Sorry for the delay, I composed a simple test to check the self-consistency of the noise patterns. Of course the previous elens tests are all still compatible with this addition.

PR #578, opened 2025-01-22
Specialise twiss

freddieknets:
## Description

This PR makes it possible to fine-tune whether or not a certain element type in the line should be included in a twiss calculation or not, using the new `skip_in_twiss` flag that can be added to a class. Regular behaviour is not affected.

In this way, one can enforce collective elements to be included in the twiss calculation (useful for instance for elements with track methods in python which still might be single-particle, hence need to be included in twiss). Example:

```python
class MyDrift:

    isthick = True
    skip_in_twiss = False

    def __init__(self, *, length=0):
        self.length = length

    def track(self, particles):
        pass
```

Similarly, one can enforce non-collective elements to NOT be included in the twiss calculation (useful for destructive elements like collimators). Example:

```python
class MyExtremeAperture(xt.BeamElement):
    skip_in_twiss = True

    _extra_c_sources = ["""
/*gpufun*/
void MyExtremeAperture_track_local_particle(DriftData el, LocalParticle* part0){
    //start_per_particle_block (part0->part)
        double y = LocalParticle_get_y(part);
        if (y > - 0.1){
            LocalParticle_set_state(part, -666);  // Die, maggots!
        }
    //end_per_particle_block
}
"""
```

## Implementation
 - When a `Tracker` is built, it will check the line for elements with non-standard behaviour and set two flags: `_skip_noncollective_elements_in_twiss` (`True` if there are some non-collective elements with `skip_in_twiss=True`) and `_enable_collective_elements_in_twiss` (`True` if there are some collective elements with `skip_in_twiss=False`). Of course, by default collective elements act as skip_in_twiss=True` and non-collective elements as `skip_in_twiss=False`). If any of these tracker flags is `True`, the tracker makes an element dict `_element_dict_for_twiss` with `Drifts` where needed (when both these flags are `False`, this dict is just equal to the regular non-collective element dict in the tracker).
 - Whenever a function needs to twiss, it will call `_get_non_collective_line` on the `Line` under consideration. This function is updated to handle the different cases. If both above flags are `False`, it works the same as before and returns itself. If this tracker is already on a line specialised for its twiss (`_specialised_for_twiss=True`) it will also return itself. If `_skip_noncollective_elements_in_twiss` is `True`, it will just make a shallow copy of the tracker and adapt its dict, just as in the regular case. However, if `_enable_collective_elements_in_twiss` is `True` it needs to completely rebuild the tracker as now some collective elements are included and hence the structure is different.
 - Because rebuilding the tracker is rarely needed but time-consuming, the specialised line is cached. The cache is invalidated if the original tracker on the original line gets invalidated and rebuilt. This is inspected by storing the memory address of the tracker of the original line into the specialised line (in `_tracker_id`). Every time a specialised line is needed, the memory addresses are compared and if invalid it is rebuilt.
 - The functions that call `_get_non_collective_line` (twiss, closed orbit search, one turn matrix, multisetter, build particles, build longitudinal particles) are adapted. Instead of first checking if the line is collective, they now directly call `_get_non_collective_line` because otherwise the case of a non-collective line with skipped elements might be missed. In any case, if the line is collective and no classes behave differently, the function does the same as before. There is a parallel PR in `xpart` with these small changes. For ease of messaging, the logging info is moved into the specialised line generation (and hence only outputted once instead of multiple times).

## Checklist

Mandatory: 

Optional:

### DIFF
--- a/xtrack/line.py
+++ b/xtrack/line.py
@@ -4477,67 +4477,82 @@ class Line:
             return self
 
         else:
-            if self.tracker._skip_noncollective_elements_in_twiss and _print:
-                log.warning(
-                    "The tracker has non-collective elements marked by "
-                    "'skip_in_twiss=True'.\nIn the twiss computation these "
-                    "elements are replaced by drifts.")
-            if self.tracker._enable_collective_elements_in_twiss and _print:
-                log.warning(
-                    "The tracker has collective elements.\n"
-                    "In the twiss computation collective elements are"
-                    " replaced by drifts (except those marked by "
-                    "'skip_in_twiss=False').")
-            elif self.iscollective and _print:
-                log.warning(
-                    'The tracker has collective elements.\n'
-                    'In the twiss computation collective elements are'
-                    ' replaced by drifts.')
+            # Return a specialised line for twiss (collective elements are replaced
+            # by drifts unless specified, and some non-collective elements as well)
+            if not hasattr(self, '_line_for_twiss') \
+            or id(self.tracker) != self._line_for_twiss._tracker_id:
+                # Rebuild the specialised line if the tracker was rebuilt
+                self._specialise_for_twiss()
+            return self._line_for_twiss
 
-            # Shallow copy of the line
-            out = Line.__new__(Line)
-            out.__dict__.update(self.__dict__)
+    def _specialise_for_twiss(self):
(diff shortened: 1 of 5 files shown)


giadarol:
Will be handled by LineTwisser class when implemented

PR #584, opened 2025-01-29
Fix: s_tol not correctly taken into account in `Line.insert(...)`

giadarol:


giadarol:
embedded in different pull request

PR #585, opened 2025-02-05
Create a dedicated `RBend` element

szymonlopaciuk:
## Description

Introduce a dedicated `RBend` to Xsuite, representing a rectangular bend. Closes xsuite/xsuite#591, as part of it becomes irrelevant, and docstrings are completed.

The `RBend` element accepts `length`, `length_straight`, `angle`, and `h` parameters (among the others shared with `Bend`) and only one of the lengths and either the angle or the curvature need to be given, with the other params being computed; if more are given, they will be checked for consistency. When changing the parameters `angle` and `length_straight` are treated as invariant: e.g., when angle is changes, the arc length and h will change, but not `length_straight`.

Caveat: there is now a difference between the way MAD-X imported `rbend`s behave between the cpymad-loader and the native madloader, when `rbarc = false` (e.g. the LHC lattice; `rbarc` is a flag, either global, or per element, that indicates whether the `rbend.l` attribute is the straight length (`rbarc`) or the arc length (`-rbarc`)). There is no setting for `rbarc` in Xsuite, so when the element is created (taking MAD's `l` as `length` or `length_straight` as required) Xsuite cannot remember which kind of an `rbend` it was: the `length_straight` of an `RBend` is the invariant in Xtrack when changing the angle, therefore discrepancies can appear once the angle is manipulated after element creation. Meanwhile, the cpymad loader loads the complete model that is aware of MAD-X state. In effect, the lengths of the rbends in this scenario can differ marginally, though Xsuite's behaviour in this case should be more "reasonable".

## Checklist

Mandatory: 

Optional:


giadarol (changes requested):
we probably need to add a `test_thick_slice_rbend` similar to `test_thick_slice_bend`

PR #596, opened 2025-02-14
Fix `RBend` radiation configuration issue in `Line.configure_radiation` method

gbrogginess:
## Description
This pull request fixes an issue in the `Line.configure_radiation` method where the `radiation_flag` attribute was being set on thick `RBend` elements, causing an `AttributeError`. The solution adds `RBend` to the list of element types that are skipped during the iteration over the `line.element_dict` when configuring radiation. This ensures that only thin elements (or allowed thick elements) have their `radiation_flag` set, preventing errors when using `RBend` elements with synchrotron radiation enabled.

## Checklist

Mandatory: 

- [] I have added tests to cover my changes

Optional:

### DIFF
--- a/xtrack/line.py
+++ b/xtrack/line.py
@@ -2992,7 +2992,7 @@ class Line:
             self._bhabha_model = None
 
         for kk, ee in self.element_dict.items():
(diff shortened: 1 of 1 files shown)


giadarol:
Solved

PR #599, opened 2025-02-18
Add support for non-deferred expressions

szymonlopaciuk:
## Description

Add true support for non-deferred expressions in the native madloader. Before `=` was always interpreted as a `:=`.

Needs xsuite/xdeps#97.

## Checklist

Mandatory: 

Optional:


giadarol:
Not relevant anymore

PR #628, opened 2025-04-14
Update trajectory_correction.py

554sj4606:
I identified a minor issue in the computation of the response matrix used for trajectory correction: where it uses the horizontal tune instead of the vertical tune when calculating the vertical response matrix.


giadarol:
Thank you! That was a biggie!

PR #631, opened 2025-04-26
feat: trajectory correction with bounds

paa27:
## Description
Added feature to use bounded least squares ([scipy implementation](https://docs.scipy.org/doc/scipy/reference/generated/scipy.optimize.lsq_linear.html)) to perform trajectory correction. If user provides corrector bounds, bounded least squares is used, instead of unbounded least squares (both for SVD and Micado).

## Checklist

Mandatory: 

- [~] All the tests are passing, including my new ones -> all tests passing on test_trajectory_correcton.py

Optional:

- [x] I have updated the docs in relation to my changes, if applicable -> updated relevant docstrings

### DIFF
--- a/tests/test_trajectory_correcton.py
+++ b/tests/test_trajectory_correcton.py
@@ -532,7 +532,7 @@ def test_orbit_correction_tilt_monitors():
     # Check that there is no vertical reading in the tilted bpm BPMs
     xo.assert_allclose(correction.y_correction._position_before, 0, atol=1e-15, rtol=0)
 
-def test_orbit_correction_with_bounds():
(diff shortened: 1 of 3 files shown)


giadarol:
Thank you for the useful feature and implementation! 

One small change request to be consistent with the nomenclature we use in match: can you call them `limits` intstead of `bounds`?

paa27:
Done :). Changes:
- Renamed `*bounds*` to `*limits*` throughout PR. 
- Updated relevant docstrings.
- Re-ran tests on test_trajectory_correcton.py -> PASSED

PR #654, opened 2025-06-16
Fix `madng_twiss` for the reference recv from `pymadng`

szymonlopaciuk:
## Description

This fixes a current issue with the madng_interface in Xsuite: pymadng now sends dictionaries as references, this adapts Xsuite to this. There is a somewhat related issue on pymadng: https://github.com/MethodicalAcceleratorDesign/MAD-NG.py/issues/24.

## Checklist

Mandatory: 

Optional:

### DIFF
--- a/xtrack/madng_interface.py
+++ b/xtrack/madng_interface.py
@@ -71,7 +71,7 @@ def _tw_ng(line, rdts=[], normal_form=True,
     mng_columns_to_send = ["mtbl." + col for col in columns]
     send_cmd = f'''
         columns = {{{", ".join(mng_columns_to_send)}}}
-        py:send(columns)
+        py:send(columns, true)
     '''
 
     if len(rdts) > 0:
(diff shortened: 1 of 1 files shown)


xtrack/madng_interface.py
```diff
@@ -159,10 +163,11 @@ def _tw_ng(line, rdts=[], normal_form=True,
                     nf:anhy{0, 1}, -- dqy / d(2 jy)
                     nf:anhx{0, 1}, -- dqx / d(2 jy)
                     nf:anhy{1, 0}, -- dqy / d(2 jx)
-                    })
+                    }
+            py:send(normal_forms_to_send)
         ''')
         mng.send(mng_script_nf)
-        out_nf = mng.recv()
```
jgray-19, line 165: I see what you mean by your comment on anonymous variables. Therefore, another fix could be to do py:send({...}, true). What could be faster is to do:
```
py:send(vector{...})
```
But this means you get a 2D numpy vector instead of a list.

PR #660, opened 2025-07-08
Inital commit

jgray-19:
## Description
Here I implement a thick and a thin AC-dipole along with tests. 

Currently fails on personal computer due to:
```
E   NameError: name 'ThickSliceSolenoid' is not defined

python3.11/site-packages/xsuite/kernel_definitions.py:72: NameError
```

Will the documentation inside the element classes be automatically inserted? I.e. Do I need to write additional documentation? 

Also, help with how to run the GPU contexts would be nice!

## Checklist

Mandatory: 

- [ ] All the tests are passing, including my new ones. 

Optional:

- [ ] I have updated the docs in relation to my changes, if applicable.
- [ ] I have tested also GPU contexts.

### DIFF
--- a/tests/test_acdipole.py
+++ b/tests/test_acdipole.py
@@ -1,0 +1,246 @@
+from collections import namedtuple
+
+import pytest
+import xpart as xp
+from xobjects.test_helpers import for_all_test_contexts
+
+import xtrack as xt
+
+
+# =====================
+# Utility Functions
+# =====================
+def _create_test_particles(at_turn: int) -> xp.Particles:
+    """Create test particles and verify initial conditions."""
+    particles = xp.Particles(at_turn=at_turn)
+    assert particles.x[0] == 0.0, f"Expected initial x=0, but got x={particles.x[0]}"
+    assert particles.y[0] == 0.0, f"Expected initial y=0, but got y={particles.y[0]}"
+    assert particles.py[0] == 0.0, (
+        f"Expected initial py=0, but got py={particles.py[0]}"
+    )
+    assert particles.px[0] == 0.0, (
+        f"Expected initial px=0, but got px={particles.px[0]}"
+    )
+    return particles
+
+
+def get_acdipole_results(
+    test_context,
+    acdipole_class: type[xt.BeamElement],
+    turn: int,
+    test_voltage: float = 1.5,
+    test_freq: float = 0.25,
+    test_lag: float = 0.0,
+) -> tuple:
+    """Track particles through an ACDipole and return coordinates."""
+    particles = _create_test_particles(at_turn=turn)
+
+    acdipole = acdipole_class(
+        volt=test_voltage,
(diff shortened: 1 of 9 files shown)


tests/test_acdipole.py
```diff
@@ -0,0 +1,246 @@
...
+        else:
+            assert vals[coord] == 0.0, (
+                f"Turn {test_turn}: Expected {coord}=0, but got {coord}={vals[coord]}"
+            )
+
+
+# =====================
+# Flattop Test Parameters and Helper
+# =====================
+flattop_cases = [
```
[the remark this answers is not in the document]
szymonlopaciuk: Also the labels can be generated on the fly maybe

szymonlopaciuk (changes requested):

tests/test_acdipole_twiss.py
```diff
@@ -0,0 +1,63 @@
...
+    line.build_tracker(_context=test_context)
+    base_tws = line.twiss(method="4d")
+    nat_qx, nat_qy = base_tws["qx"], base_tws["qy"]
+    e5_pos = line.get_s_position("e5")
+    e5_betx = base_tws.rows["e5"]["betx"].item()
+    e5_bety = base_tws.rows["e5"]["bety"].item()
+
+    drv_qx, drv_qy = nat_qx + qx_shift, nat_qy + qy_shift
+
+    # Copy the line before inserting elements
```
szymonlopaciuk: Outdated comment

xtrack/beam_elements/acdipole.py
```diff
@@ -0,0 +1,221 @@
...
+            natural_q = 0
+
+        if not (
+            (isinstance(ramp, (list, tuple)) or hasattr(ramp, "__iter__"))
+            and len(ramp) == 4
+            and all(isinstance(int(x), int) for x in ramp)
+        ):
+            raise ValueError(
+                "The ramp parameter must be a list of four integers: [ramp1, ramp2, ramp3, ramp4]."
+            )
```
szymonlopaciuk: I find this a bit complicated for what it does.

1. `isinstance(ramp, (list, tuple))` implies `hasattr(ramp, "__iter__")`, so the first check is not necessary
2. In most cases `__iter__` implies a `__len__`. Sure there are exceptions, but on the other hand it's not strict enough anyway: `set` passes all the checks, but it's probably not desired. You'd probably also need to add a check to assert that the list is sorted as well?
3. Is the check for integer not a bit strict? What if someone gives a float for whatever reason?

I agree the built-in error message in Xobjects is not the clearest, so having some sort of check is an improvement for the user. I'm not sure what is the best way, but I'd try to simplify. One possibility:

```python
if len(ramp) != 4 or list(ramp) != sorted(ramp):
    raise ValueError(
        "The ramp parameter must be an increasing list of four integers."
    ) 
```

PR #665, opened 2025-07-22
Quadrupole wedge

silkevdschueren:
## Description

Quadrupole wedge as in PTC with option MAD8_WEDGE=False


silkevdschueren:
Replaced by #670.

PR #670, opened 2025-07-29
Quadrupole wedge

silkevdschueren:
## Description

Quadrupole wedge as in PTC with option MAD8_WEDGE=False.
So far only with hardcoded test for quadrupole wedge with hard edge quadrupole fringe. It would be good to compare with PTC (option MAD8_wedge=False), or with Hamiltonian solver if quadrupole wedge element can be isolated.

Still in progress.

This PR replaces  PR #665 (as it was from the fork's main).

## Checklist

Mandatory: 

Optional:


giadarol:
Merged manually

PR #688, opened 2025-08-31
Improved atomic add

freddieknets:
## Description

This PR updates the `atomicAdd` implementation to be compatible with the new API headers and includes, and makes it type-safe on any context, CPU or GPU.

In particular, the following function definitions are added:

```C
int8_t   atomicAdd_i8(GPUGLMEM int8_t* addr, int8_t val);
int16_t  atomicAdd_i16(GPUGLMEM int16_t* addr, int16_t val);
int32_t  atomicAdd_i32(GPUGLMEM int32_t* addr, int32_t val);
int64_t  atomicAdd_i64(GPUGLMEM int64_t* addr, int64_t val);
uint8_t  atomicAdd_u8(GPUGLMEM uint8_t* addr, uint8_t val);
uint16_t atomicAdd_u16(GPUGLMEM uint16_t* addr, uint16_t val);
uint32_t atomicAdd_u32(GPUGLMEM uint32_t* addr, uint32_t val);
uint64_t atomicAdd_u64(GPUGLMEM uint64_t* addr, uint64_t val);
float    atomicAdd_f32(GPUGLMEM float* addr, float val);
double   atomicAdd_f64(GPUGLMEM double* addr, double val);
```
which implements typed versions of `atomicAdd` forwarding to the relevant built-in if it exists, or defining it based on the atomic Compare-And-Swap functions of the GPU compilers, taking the compiler version into account if needed.

Furthermore, a generic overloaded function is defined that auto-recognises the type and forwards to the relevant definition above:

```C
<...> atomicAdd(GPUGLMEM <...>* addr, <...> val);
```

The overloading behaviour depends on `_Generic` (from C11 onwards) for CPU contexts, on `extern "C++"` for CUDA contexts (always fine as CUDA is implemented in C++), and on `__attribute__((overloadable))` for OpenCL contexts (which is a `clang` attribute, but it seems `pyopencl` always supports it).

A dedicated test is added, which tests for concurrency issues for all supported types.

This has been tested on an NVIDIA 1050 Ti for all contexts (CPU, OpenMP, CUDA, OpenCL).

## Checklist

Mandatory: 

Optional:

### DIFF
--- a/xtrack/headers/atomicadd.h
+++ b/xtrack/headers/atomicadd.h
@@ -142,7 +142,7 @@ DEF_ATOMIC_ADD(double  , f64)
     #pragma OPENCL EXTENSION cl_khr_int64_base_atomics : enable
     #pragma OPENCL EXTENSION cl_khr_fp64 : enable
     #define __XT_CAS_U32(ptr, exp, val) atomic_cmpxchg((GPUVOLATILE uint32_t*)(ptr), (uint32_t)(exp), (uint32_t)(val))
-    #define __XT_CAS_U64(ptr, exp, val) atomic_cmpxchg((GPUVOLATILE uint64_t*)(ptr), (uint64_t)(exp), (uint64_t)(val))
+    #define __XT_CAS_U64(ptr, exp, val) atom_cmpxchg((GPUVOLATILE uint64_t*)(ptr), (uint64_t)(exp), (uint64_t)(val))
     #define __XT_AS_U32_FROM_F32(x) as_uint(x)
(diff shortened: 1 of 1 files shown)


szymonlopaciuk (approved):
Looks good to me, and I think it's useful to have a uniform set of atomic adds available (especially considering the horrors of the different OpenCL function versions we've seen yesterday with Frederik...)

I have verified there is no visible performance difference either in compilation or in runtime with this PR. No issues on AMD either with the latest fixes. I think we can proceed.

Perhaps putting this in Xobjects at some point could be appropriate, but I'm happy to merge this now.

PR #710, opened 2025-11-05
TPSA in Xsuite Matching (Interface to MAD-NG)

babreufig:
## Description
Implements MAD-NG TPSA Action to use MAD-NG tracked damaps to acquire optical parameters.

This pull request requires at least MAD-NG version 1.1.8. There is a discrepancy in the chromaticity calculation between MAD-NG version 1.1.7 and 1.1.8, where the `phix` is by a factor 2 smaller in the new version. This seems to be a different convention(?). Needs a follow-up because this affects Ax/Bx/Ay/By functions which are then different between Xsuite and MAD-NG. For reference: https://github.com/MethodicalAcceleratorDesign/MAD-NG/commit/5c87e03e1cabe4aa01a4581370edb31e747558b4

Given the revert from the chromaticity bug, this pull request should be shipped with MAD-NG version 1.1.9!

**Important:** Depends on https://github.com/xsuite/xdeps/pull/106

- [x] Supports ranged settings with initial conditions
- [x] Supports full-lattice settings with and without initial conditions
- [x] Supports Beta functions, Alpha functions, Dispersions and Phase Advances as Targets
- [x] Phase Advances uses faster Twiss and TPSA as well
- [x] Interchangability between Xsuite ('betx', 'bety') and MAD-NG ('beta11_ng', 'beta22_ng') target names
- [x] Orbit targets supported (x, px, ...)
- [x] Changed twiss function to put coupling and chromaticities as optional flags
- [x] Support R-Matrix matching 
- [x] Conversion between zeta/t and delta/pt
- [x] Added multipolar components to Sextupoles and Octupoles for Xsuite-to-MAD-NG conversion (knl/ksl)
- [x] Added longitudinal shift for Xsuite-to-MAD-NG conversion (shift_s) and rot_s_frame_no_frame (dpsi)

Mandatory: 


xtrack/madng_interface.py
```diff
@@ -15,22 +15,129 @@
...
+        # Check for key if it's a ctpsa or tpsa
+        var = f"MADX['{key.replace('.', '_')}']"
+        is_tpsa = self.mad.send(f"py:send(MAD.typeid.is_tpsa({var}) or MAD.typeid.is_ctpsa({var}))").recv()
+        if is_tpsa:
+            self.mad.send(f"{var}:set0(py:recv())").send(value)
+        else:
+            self.mad[var] = value
+
+
         #Expressions still to be handled, could use the following:
```
szymonlopaciuk: Maybe can open an issue for this if it's not to be part of this PR.

xtrack/madng_interface.py
```diff
@@ -93,25 +200,31 @@ def _build_rdt_script(mng_sequence_name, rdts, columns):
 def _build_beta0_block_string(tw_kwargs):
     flag_init = False
     beta0_keys = []
+    beta0_vals = []
```
szymonlopaciuk: At this point this could become a dictionary (:

xtrack/madng_interface.py
```diff
@@ -135,33 +248,35 @@ def _tw_ng(line, rdts=(), normal_form=True,
             raise NotImplementedError('TwissTable as init not implemented.')
         X0_str = _build_beta0_block_string(tw_kwargs)
     else:
-        X0_str = f'X0={X0}, '
+        X0_str = f'X0 = {X0}, '
+
+    if start is not None and end is None or start is None and end is not None:
```
szymonlopaciuk: Maybe more readable and shorter with `(start is None) != (end is None)`?

xtrack/madng_interface.py
```diff
@@ -391,11 +515,481 @@ def prepare(self, force=False):
...
+        self.match_opt = False
+        self.sum_rmat_tar = sum_rmat_tar
+        self.rmat_start_end_list = None
+        self.rmat_tags = None
+        self._last_res = None
+        self._needs_zeta_scale = []
+        self._needs_delta_scale = []
+
+    def prepare(self, force=False):
+        '''
```
szymonlopaciuk: Sorry to be picky, since I know we have this everywhere, but generally docstrings are recommended to use `"""` instead of `'''` ([PEP 257 -- Docstring Conventions](http://www.python.org/dev/peps/pep-0257/)).

xtrack/madng_interface.py
```diff
@@ -391,11 +515,481 @@ def prepare(self, force=False):
...
+
+                    self.rmat_start_end_list[idx] = (loc_start, loc_end)
+                    self.rmat_tags.append(target.tag)
+
+                    targets_map_str += f"{XSUITE_MADNG_ENV_NAME}.targets_arr[{i+1}] = {{ loc = '{loc_end}', qty = '{qty}', loc_start = '{loc_start}', rmat = true, tag = {tag} }}\n"
+                    xs_ng_target_map += f"{XSUITE_MADNG_ENV_NAME}.xs_ng_target_map['{target.term}'] = '{qty}'\n"
+
+                else:
+                    raise NotImplementedError(f"Target of type {type(target)} not implemented for MAD-NG TPSA matching.")
+
```
szymonlopaciuk: There is no `else:`, which means if someone provides a broken/wrong target, the error will pass silently. What is the expected type in l. 658? I assume `Target`, a subclass, or some other compatible ducktyped construct, in which case, maybe just change that line to `else:` and let whatever errors happen in that branch, in case the user passed a wrong thing.

There's some repeated code between the branches.

Also as discussed irl, at some point let's discuss the interface of `Target`.

xtrack/madng_interface.py
```diff
@@ -391,11 +515,481 @@ def prepare(self, force=False):
...
+        self.mng.send(mng_init_str)
+
+        self._already_prepared = True
+
+    def run(self):
+        '''
+        Execute the MAD-NG TPSA matching action.
+        This method performs either a Twiss or Track operation in MAD-NG
+        depending if quantities can be calculated through tracking or not.
```
szymonlopaciuk: See my later comment of the `twiss_flag`. Technically, here you "perform twiss or track operation in MAD-NG depending" on the `self.twiss_flag`, the logic for the quantities is external to the class, and as far as the action is concerned it doesn't matter. (My later comment: maybe this logic could be pulled in? Not sure if it makes sense, you have the big picture so you will know best.)

xtrack/madng_interface.py
```diff
@@ -391,11 +515,481 @@ def prepare(self, force=False):
...
+            f"}}\n"
+            f"{XSUITE_MADNG_ENV_NAME}.trk = trk\n"
+        )
+
+        self.mng.send(mng_track_str)
+
+        loc_map_str = ''
+        loc_map_str = "\n".join(f"{XSUITE_MADNG_ENV_NAME}.target_loc_map['{loc}'] = {XSUITE_MADNG_ENV_NAME}.trk['{loc}'].__map" for loc in self.optics_target_locations)
+
+        # ===== TWISS =====
```
szymonlopaciuk: Same comment as earlier. Maybe the branches can be extracted to methods?

xtrack/match.py
```diff
@@ -607,6 +608,9 @@ def compute(self, tw):
             'Only terms of the R-matrix in the form "r11", "r12", "r21", "r22", etc'
             ' are supported')
 
+        if hasattr(tw, '_data') and hasattr(tw._data, 'attrs') and self.tag in tw._data.attrs:
```
szymonlopaciuk: Doesn't `tw` always have `_data`? 😮

szymonlopaciuk (changes requested):
Good job, it's quite well structured and readable!

Some requests/comments/questions/suggestions for your consideration.

xtrack/madng_interface.py
```diff
@@ -15,22 +15,129 @@
...
+TPSA_ALLOWED_TARGETS = { 'beta11', 'beta22', 'alfa11', 'alfa22', 'dx', 'dpx', 'dy', 'dpy',
+                         'mu1', 'mu2', 'x', 'px', 'y', 'py', 't', 'pt' }
+
+XSUITE_MADNG_ENV_NAME = "_xsuite_matching_env"
+
+def _lua_list(strings):
+    """Convert Python list → Lua { 'a', 'b', 'c' }."""
+    return "{ " + ", ".join(f"'{s}'" for s in strings) + " }"
+
+def dp2pt(dp, beta0):
```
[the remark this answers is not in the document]
babreufig: I used it before I started using ptau, so now it's not needed anymore. I'll remove

xtrack/madng_interface.py
```diff
@@ -15,22 +15,129 @@
...
+    Returns
+    -------
+    float
+        Transverse momentum relative to total momentum (pt/p, dimensionless).
+    """
+
+    _beta0 = 1 / beta0
+    return np.sqrt((1 + dp) ** 2 + (_beta0**2 - 1)) - _beta0
+
+def pt2dp(pt, beta0):
```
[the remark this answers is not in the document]
babreufig: You mean as a C kernel? I don't see that it's used that often, I only use it once to convert the MAD-NG particles to Xsuite ones. In Xsuite we support both delta and ptau, so..

xtrack/madng_interface.py
```diff
@@ -360,10 +295,10 @@ def _process_data(data):
         tw[nn] = np.atleast_1d(np.squeeze(out_dct[nn]))[:-1]
 
     if compute_chromatic_properties:
-        temp_x = tw.wx_ng * np.exp(1j*2*np.pi*tw.phix_ng)
+        temp_x = tw.wx_ng * np.exp(1j*2*np.pi*tw.phix_ng*2)
```
szymonlopaciuk: I didn't request it, but it changed.

szymonlopaciuk (approved):
Good for me, just the little comment.

PR #727, opened 2025-11-24
Add exit aperture after slicing.

freddieknets:
## Description

When slicing, apertures are automatically added between the slices if the element has `name_associated_aperture` set. However, no aperture is set after the last slice (which can be problematic when backtracking e.g. during aperture interpolation). This PR adds a final aperture at the end (before the `exit_marker` if present).

## Checklist

Mandatory: 

Optional:


giadarol:
deployed

PR #741, opened 2025-12-19
Fix k0 property to return actual value from h

eltos:
## Description

In recent versions, [`Bend.k0`](https://github.com/xsuite/xtrack/blob/fe5788f99cd76619659a86224a38f43c0d28eb7d/xtrack/beam_elements/elements.py#L1446) may return `"from_h"` as a string. Getting a string where a float is expected likely breaks a lot of code, and requires users to always check for this special string to get the actual value. It's also not necessary, given that [`Bend.k0_from_h`](https://github.com/xsuite/xtrack/blob/fe5788f99cd76619659a86224a38f43c0d28eb7d/xtrack/beam_elements/elements.py#L1464) exists and can be used to determine if `k0` is set manually or determined by `h`.

Not sure what the rationale behind returning a special string here was, but I propose returning the actual value (the value of `h`) in such cases.

Also fixes a missing assignment of k0 if h is set to zero

Related: https://github.com/xsuite/xplt/commit/31581073a41d5f617f6bb03c4d466bb04182d65d

## Checklist

Mandatory: 

Optional:

### DIFF
--- a/xtrack/beam_elements/elements.py
+++ b/xtrack/beam_elements/elements.py
@@ -1428,10 +1428,10 @@ class _BendCommon(_HasKnlKsl, _HasIntegrator, _HasModelCurved):
         self._length = value
         if self.length != 0:
             self._h = self.angle / self.length
-            if self.k0_from_h:
-                self._k0 = self.h
         else:
             self._h = 0.0
+        if self.k0_from_h:
+            self._k0 = self.h
 
     @property
     def h(self):


eltos:
A possible alternative solution, if you are interested, would be the following:
If we clearly distinguish between the internal state (`_k0`) and user exposed properties (`k0`), inconsistencies can be strictly avoided and the change of issues reduced. Cloning, serialising and deserialising objects is then done using only the internal state variables (ideally with proper clone, to_dict, from_dict like methods across the whole class hierarchy with proper super calls). Users on the other hand should touch the exposed properties to alter the object state consistently. And if a user changes them, then it makes sense that the order matters (as in your example where `k0` is set to an actual value), which is the present behaviour already now.

Such a refactoring will probably require a larger effort, let me know if you want to further discuss it or not.

giadarol:
> A possible alternative solution, if you are interested, would be the following: If we clearly distinguish between the internal state (`_k0`) and user exposed properties (`k0`), inconsistencies can be strictly avoided and the change of issues reduced. Cloning, serialising and deserialising objects is then done using only the internal state variables (ideally with proper clone, to_dict, from_dict like methods across the whole class hierarchy with proper super calls). Users on the other hand should touch the exposed properties to alter the object state consistently. And if a user changes them, then it makes sense that the order matters (as in your example where `k0` is set to an actual value), which is the present behaviour already now.
> 
> Such a refactoring will probably require a larger effort, let me know if you want to further discuss it or not.

The issue is that when we clone we need to preserve the xdeps expressions (which in general creates a lot of constraints but are a must to handle large machines). The expressions control user-exposed properties and not internal state

PR #752, opened 2026-02-16
Add coupled transverse motion to LineSegmentMap with Cij matrix

betelgeuse1703:
## Description
Adding coupling in line-segment maps with Cij matrix, 

f1001 and f1010 from twiss matches the input rdts used for generating four elements of Cij


betelgeuse1703:
This PR fixes the sign convention in the coupled LineSegmentMap transport matrix.

- Correct sign in coupling matrix 
-  test for coupled optics transport through LineSegmentMap (test_coupled_line_segment_map)
- Remove accidental AFS temporary files committed previously

This change only affects coupled transport in LineSegmentMap.
input rdts to compute coupling matrix matches output rdts from twiss

giadarol:
Can you please expand the PR description with the formulas that have been implemented and some references to relevant papers?

PR #753, opened 2026-02-19
Aperture modelling tools for Xtrack

szymonlopaciuk:
## Description

First prototype of the aperture modelling functionality. See the examples for usage and the associated physics manual PR xsuite/xsuite#809.

Needs: xsuite/xobjects#169.

Currently allows:

- Loading an aperture model from a line with limits, a line with MAD-X metadata, or a line with `associated_apertures` defined on elements.
- Calculation of the n1 using different methods: by bisecting the beam envelope on the aperture polygon (`bisection`), using a linear-complexity approximated `rays` method, and a more precise quadratic-in-rays `exact` method.
- Calculation of the beam envelope at arbitrary sigmas and pipe cross sections.
- Can interpolate cross sections, even in curved, arbitrarily transformed types, with arbitrarily transformed profiles.

Caveats:
- GPU support not yet implemented.
- Interpolation along curved types uses an approximation (that should be "good enough" but should be kept in mind).

## Checklist

Mandatory: 

Optional:


szymonlopaciuk:
Already merged squashed

PR #782, opened 2026-05-19
Removed CUDA context from ACDipole Test

babreufig:
## Description
Removed CUDA Context for ACDipole test as MAD-NG doesn't use GPU's and multi-turn tracking with monitor is not supported in the GPU context.


szymonlopaciuk:
I think we have this (in a slightly different way https://github.com/xsuite/xtrack/blob/main/tests/test_acdipole.py#L265). Feel free to reopen if needed.

PR #790, opened 2026-06-04
Enable RandomRutherford on the CUDA (Cupy) GPU context

InnovateInfinity87:
# Enable `RandomRutherford` on the CUDA (Cupy) GPU context

## Summary

`RandomRutherford` was blocked from all non-CPU contexts by a Python guard in
`RandomRutherford.__init__`, even though its C sampling kernel is already
GPU-clean (`GPUFUN`/`GPUKERN`, `#ifdef XO_CONTEXT_CPU` include guards, a
fixed-count Newton inversion, no `malloc`/`printf`/global RNG state). This
narrows the guard to reject only OpenCL — the file-scope `static const double
ei[]` table in `exponential_integral_Ei.h` is illegal in OpenCL but fine on CUDA
(constant memory) — which enables the Rutherford RNG on the Cupy/CUDA context.

## What changed

- `xtrack/random/random_generators.py`: replace the blanket
  `if not isinstance(self._context, xo.ContextCpu): raise ...` guard with a
  narrow `if isinstance(self._context, xo.ContextPyopencl): raise ...` — CUDA/Cupy
  is now allowed; OpenCL is still rejected with a clear message.
- `tests/test_random_gen_ruth.py`: stop excluding `ContextCupy` from the existing
  Rutherford tests; add `test_cpu_gpu_distribution_match` (a CPU-vs-GPU
  distributional check).
- `examples/random_number_generator/002_rutherford_cpu_vs_gpu.py`: a CPU-vs-GPU
  comparison figure generator (log-y PDF overlay against the analytic Rutherford,
  empirical CDFs/KS, and the GPU/CPU bin-ratio vs a counting-noise band).

## Validation

- `pytest tests/test_random_gen_ruth.py` → **7 passed**: all three existing tests
  now run on **both** `ContextCpu` and `ContextCupy` (analytic-PDF agreement and
  within-context reproducibility hold on GPU), plus the new cross-context test.
- CPU-vs-GPU agreement on 3e6 samples: **KS = 0.0006** (p ≈ 0.6), **KL ≈ 5.6e-5**
  (thresholds KS < 0.01, KL < 1e-3). Means/stds match to ~4 significant figures.

Agreement is distributional, not per-particle bitwise: the Newton + Ei series
need only converge to within sampling noise, and the parallel RNG streams differ
across contexts.

## Notes

- OpenCL remains out of scope (the file-scope `ei[]` table is OpenCL-illegal);
  CUDA/Cupy only.
- This unblocks a follow-up `xcoll` PR that makes `EverestCrystal` GPU-trackable:
  it holds a `RandomRutherford` field and calls `RandomRutherford_generate` in its
  nuclear-interaction path, so it cannot instantiate on CUDA without this change.

### DIFF
--- a/examples/random_number_generator/002_rutherford_cpu_vs_gpu.py
+++ b/examples/random_number_generator/002_rutherford_cpu_vs_gpu.py
@@ -1,0 +1,118 @@
+"""Visual CPU-vs-GPU comparison of the RandomRutherford sampler.
+
+Samples the Rutherford distribution on ContextCpu and ContextCupy with identical
+parameters and overlays the two distributions against the analytic PDF. This is
+the documentation figure for enabling RandomRutherford on the CUDA context:
+agreement is distributional (parallel RNG streams + FP ordering differ across
+contexts, so a per-particle bitwise match is neither expected nor required).
+
+Run with cupy's CUDA libs on the loader path, e.g.:
+    LDP=$(ls -d $(python -c 'import sys;print(sys.prefix)')/lib/python*/site-packages/nvidia/*/lib | tr '\n' ':')
+    LD_LIBRARY_PATH="$LDP" python 002_rutherford_cpu_vs_gpu.py
+"""
+import numpy as np
+import matplotlib.pyplot as plt
+from scipy.stats import ks_2samp
+
+import xobjects as xo
+import xtrack as xt
+
+# Rutherford parameters (same as tests/test_random_gen_ruth.py)
+rA, rB, t0, t1, iterations = 0.0012306225579197868, 53.50625, 0.001, 0.02, 7
+N_SAMPLES, N_SEEDS = int(3e6), 3000
+
+
+def sample(context):
+    ran = xt.RandomRutherford(_context=context)
+    ran.A, ran.B = rA, rB
+    ran.lower_val, ran.upper_val = t0, t1
+    ran.Newton_iterations = iterations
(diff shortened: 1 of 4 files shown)


freddieknets:
Have you looked at the timing of the GPU runs? I remember that one of the reasons to exclude the GPU context was that, due to the way the Ei table was constructed and loaded, it ended up being literally slower than on CPU.

As for your upcoming Xcoll PR, feel free to involve me in advance. There are other reasons to limit Everest to CPU, like the use of function pointers in the geometry checks. Also, we are working on a big rewrite of Everest (more correct physics, but also better geometry, and adding GPU support), so we can have an early look to avoid future incompatibilities.

InnovateInfinity87:
Not at the RS directly but at the whole implementation of the crystal. It looks promising, but I did not spend any time in optimizing that (or even looking into it). This is just the changes needed to make it work. If you are happy with this first PR, then I can also open the next once this is approved. 

Thank you very much for looking at this and for making the first version of this - it is really incredibly useful! 

everest_crystal_gpu_benchmark

InnovateInfinity87:
And here you have the comparison for only RS. 

rutherford_cpu_vs_gpu_benchmark

freddieknets (approved):
Fine with me to merge. The speed gain from moving to GPU seems a bit on the low side, but that should not be a reason to block the merge. We can always optimise later, if possible.

InnovateInfinity87:
great thanks! Indeed, no I did no work to try to speed that up but just to make it running with least number of changes.

PR #797, opened 2026-06-12
Integrator for curved and longitudinally dependent fields

silkevdschueren:
## Description

New beam element `FieldExpansion` to perform tracking through fields with longitudinal dependence and/or curvature based on a recursive representation ([proceedings IPAC24](https://proceedings.jacow.org/ipac2024/doi/jacow-ipac2024-tups09/)) with a Runge-Kutta integrator. The practical implementation is a factory: when creating a `FieldExpansion` instance, it spits out a `BentFieldExpansion` or `StraightFieldExpansion` depending on curvature.

Preliminary timing results (on my simple PC): tracking through a `FieldExpansion` with 100 steps takes 200us per particle per turn. Twissing a FODO cell made of `FieldExpansion`s takes similar times as a FODO cell made of regular Xsuite elements (difficult to quantify exactly due to compilation and cached twisses).

Three dedicated tests are added to the PR. Full tests of xtrack have not yet been ran completely.
Important: there is a weird issue with the last test (test_twiss). When ran sequentially, it comes last and passes without problem. But when ran in parallel using xdist, that specific test fails the twiss with a ClosedOrbitSearchError....

## Checklist

Mandatory: 

Optional:


rdemaria:
Few additional steps for Leir and Polarizer Ring use case:
- Physics manual entry
- Test backtracking
- Classes name, merging classes
- Spin and radiation
- Bs and other quantities in the table for integral

Additionally
- Compatible with RDTs (knowing that also solenoid is not supported)

rdemaria:
Here a the list of things to do:
- [ ] Rebase
- [ ] Rename Currently FieldExpansion to BFieldExpansion
- [ ] Survey Write angle to Xobjects, Ensure read only (h must write the angle)
- [ ] Review switching default to angle for construction, as is done for bend natively
- [ ] Add guards for radiation and spin
- [ ] Slicing
- [ ] Add guard to freeze things after creation because of caching
Ensure env.new and env.set work, but env.h = new_val raises
- [ ]  Adjust naming of components  b1 becomes k0 etc, Check if this fixes Twiss plotting, Provide sensible knl ksl to twiss (integral needs to be calculated
- [ ]  Additional tests, Including on GPU
- [ ] Global Xsuite issues: Decide on approach to ks in line table (currently all else are k0l, k1l, but just ks. Don't want ksl, need to decide if this is global refactor to ksol and ksoll)

PR #821, opened 2026-07-14
Fixes for the SPS aperture model import from LDB

szymonlopaciuk:
## Description

Add a complete example for the SPS aperture model imported from LDB, including the comparison of the built model and the round trip through a line with limits.

Remove the `inner_radius_sq` field from `LimitPolygon` and instead do the calculation in C directly, but only on the CPU contexts (no benefit on GPU).

Fix bug in the generation in `Aperture.get_limit_elements`.

## Checklist

Mandatory: 

Optional:


szymonlopaciuk:
Merged

PR #824, opened 2026-07-17
Handle empty MAD-X apertures

### LINKED ISSUE #877: Unable to parse sequence from MAD-X save using native loader
[mylhc.seq.zip](https://github.com/user-attachments/files/29896555/mylhc.seq.zip)

szymonlopaciuk:
## Description

Fix a problem in the native madloader, where elements with aperture params but no actual apertures would cause an error. This is malformed, however MAD-X can output these in `save`, so we must support these.

Closes xsuite/xsuite#877.

## Checklist

Mandatory: 

Optional:

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -65,3 +65,6 @@ include = "xtrack"
 markers = [
     "context_dependent: marks test as one that depends on the execution context",
 ]
+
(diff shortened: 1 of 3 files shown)


szymonlopaciuk:
Merged

PR #829, opened 2026-07-23
Allow SecondOrderTaylorMap in Environment.new(), and fix expressions in multi-dimensional array fields

JPTS2:
## Summary

Adding `SecondOrderTaylorMap` to `Environment.new()` surfaced a
pre-existing bug in how array-valued kwargs are parsed. Both are fixed
here, as the first is not usable without the second.

### The feature
- Add `xt.SecondOrderTaylorMap` to `_ALLOWED_ELEMENT_TYPES_IN_NEW`, so it
  can be built with `env.new(name, xt.SecondOrderTaylorMap, ...)` like
  `FirstOrderTaylorMap` already can.
- Give `k`/`R`/`T` defaults (zero, identity, zero), mirroring
  `FirstOrderTaylorMap`'s `m0`/`m1`, so the element defaults to an
  identity map. Documented on both classes, as the identity default was
  previously undocumented on `FirstOrderTaylorMap` too.
- Remove a duplicate `xt.XRotation` entry noticed in the same list.

### The bug
A non-zero default (`R` must default to `np.eye(6)`) can only be declared
by wrapping the field: `xo.Field(xo.Float64[6, 6], default=...)`. But
`_parse_kwargs` tested `xo.array.is_array(field)`, which returns `False`
for a wrapped array type, so the expression-parsing branch was skipped
entirely and string kwargs reached the constructor unevaluated.

Separately, that branch only ever walked one level deep, so an expression
nested inside a 2D/3D array (`R[2][3] = '2*a'`) was never evaluated at any
point in the past.

Both are fixed: `_parse_kwargs` unwraps `xo.Field`, and parsing is now
recursive to arbitrary depth via `_parse_array_value`.

Affects only fields declared with a non-trivial default —
`SecondOrderTaylorMap.k/R/T`, `FirstOrderTaylorMap.m0/m1`,
`SimpleThinQuadrupole.knl`, `Elens.coefficients_polynomial`. Ordinary
`knl`/`ksl` on magnets use bare array types and were never affected.

## Test plan
- [x] New `test_env_new_multi_dimensional_arrays` /
      `test_env_set_multi_dimensional_arrays`: expressions at 1D/2D/3D,
      live updates, and stale expressions cleared on overwrite.
- [x] New `test_env_new_whole_array_reference`: passing a whole array as a
      single reference still wires item by item (this regressed during
      development and was not covered by any existing test).
- [x] New `test_env_new_allowed_elements`: every entry in
      `_ALLOWED_ELEMENT_TYPES_IN_NEW` is constructible via `env.new`.
      Roughly half the list, including both Taylor maps, had no such
      coverage.
- [x] Full suite: 2226 passed, 3 skipped.

### DIFF
--- a/tests/test_environment.py
+++ b/tests/test_environment.py
@@ -1681,6 +1681,80 @@ def test_env_new_allowed_elements(cls_name):
     assert isinstance(env['e'], cls)
 
 
+def test_env_new_multi_dimensional_arrays():
+    # env.new() must wire expressions nested at any depth. SecondOrderTaylorMap
+    # is used because its k/R/T are 1D/2D/3D array fields.
+    env = xt.Environment()
+    env['a'] = 2.
+
+    k = [0, 0, 0, 0, 0, 0]
+    k[1] = '3*a'
+    env.new('tm_k', xt.SecondOrderTaylorMap, k=k)
+    assert env['tm_k'].k[1] == 6.
+    assert env.ref['tm_k'].k[1].xdeps.expr == "(3.0 * vars['a'])"
+    env['a'] = 5.
+    assert env['tm_k'].k[1] == 15.
+    env['a'] = 2.
+
+    R = np.eye(6).tolist()
+    R[2][3] = '2*a'
+    env.new('tm_R', xt.SecondOrderTaylorMap, R=R)
+    expected_R = np.eye(6)
+    expected_R[2, 3] = 4.
+    xo.assert_allclose(env['tm_R'].R, expected_R, atol=0, rtol=0)
+    assert env.ref['tm_R'].R[2, 3].xdeps.expr == "(2.0 * vars['a'])"
+    env['a'] = 5.
+    assert env['tm_R'].R[2, 3] == 10.
+    env['a'] = 2.
+
+    T = np.zeros((6, 6, 6)).tolist()
+    T[1][2][3] = '4*a'
+    env.new('tm_T', xt.SecondOrderTaylorMap, T=T)
+    expected_T = np.zeros((6, 6, 6))
+    expected_T[1, 2, 3] = 8.
+    xo.assert_allclose(env['tm_T'].T, expected_T, atol=0, rtol=0)
+    assert env.ref['tm_T'].T[1, 2, 3].xdeps.expr == "(4.0 * vars['a'])"
+    env['a'] = 3.
+    assert env['tm_T'].T[1, 2, 3] == 12.
+
+
+def test_env_set_multi_dimensional_arrays():
+    # env.set() must wire a new expression, and clear a stale one when an
+    # expression-bearing cell is overwritten with a plain value.
+    env = xt.Environment()
+    env['a'] = 2.
+    env.new('tm', xt.SecondOrderTaylorMap)
+
+    env.set('tm', k=['3*a', 0, 0, 0, 0, 0])
+    assert env['tm'].k[0] == 6.
+
+    env.set('tm', k=[99, 0, 0, 0, 0, 0])
+    env['a'] = 100.
+    assert env['tm'].k[0] == 99., 'stale expression on k[0] should have been cleared'
+    env['a'] = 2.
+
+    R = np.eye(6).tolist()
+    R[1][4] = '5*a'
+    env.set('tm', R=R)
+    assert env['tm'].R[1, 4] == 10.
+
+    env.set('tm', R=np.eye(6).tolist())
+    env['a'] = 50.
+    assert env['tm'].R[1, 4] == 0., 'stale expression on R[1,4] should have been cleared'
+
+
+def test_env_new_whole_array_reference():
+    # Passing a whole array as a single reference wires each item to the
+    # corresponding item of the referenced array.
+    env = xt.Environment()
+    env.new('src', xt.Quadrupole, knl=[1, 2, 3])
+    env.new('dst', xt.Quadrupole, knl=env.ref['src'].knl)
+
+    assert env['dst'].knl[0] == 1.
+    env.set('src', knl=[9, 9, 9])
+    assert env['dst'].knl[0] == 9.
+
+
 def test_env_new_prototype_keyword_and_deprecated_parent():
 
     env = xt.Environment()
--- a/tests/test_environment.py
+++ b/tests/test_environment.py
@@ -1743,6 +1743,17 @@ def test_env_set_multi_dimensional_arrays():
     assert env['tm'].R[1, 4] == 0., 'stale expression on R[1,4] should have been cleared'
 
 
+def test_env_new_array_field_errors():
+    # Bad values for an array field must be reported against that field.
+    env = xt.Environment()
+
+    with pytest.raises(TypeError, match='knl should be an iterable'):
+        env.new('q', xt.Quadrupole, length=1, knl=5)
+
+    with pytest.raises(ValueError, match='R must be a rectangular array'):
+        env.new('tm', xt.SecondOrderTaylorMap, R=[[1., 2.], [3.]])
+
+
 def test_env_new_whole_array_reference():
     # Passing a whole array as a single reference wires each item to the
     # corresponding item of the referenced array.
--- a/tests/test_environment.py
+++ b/tests/test_environment.py
@@ -1663,6 +1663,24 @@ def test_env_new():
     assert env[ret].b == 3
 
 
+@pytest.mark.parametrize('cls_name', sorted(xt.line._ALLOWED_ELEMENT_TYPES_DICT.keys()))
+def test_env_new_allowed_elements(cls_name):
+    # Every class in `_ALLOWED_ELEMENT_TYPES_IN_NEW` must be constructible via `env.new(name, cls)`.
+    # If there are mandatory arguments, they must be special-cased below.
+    cls = xt.line._ALLOWED_ELEMENT_TYPES_DICT[cls_name]
+    env = xt.Environment()
+
+    if cls_name == 'Replica':
+        env.new('base', xt.Marker)
+        env.new('e', 'base', mode='replica')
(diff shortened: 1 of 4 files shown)


JPTS2:
@giadarol was there any reason SecondOrderTaylorMap was previously excluded that I have missed?

tests/test_environment.py
```diff
@@ -1663,6 +1663,24 @@ def test_env_new():
     assert env[ret].b == 3
 
 
+@pytest.mark.parametrize('cls_name', sorted(xt.line._ALLOWED_ELEMENT_TYPES_DICT.keys()))
+def test_env_new_allowed_elements(cls_name):
```
szymonlopaciuk, line 1667: You wrote about the testing of other elements in `Environment.new` that

> This was previously untested for roughly half the list
(including both Taylor map classes) — it's now a regression guard
against any future addition breaking construction via the
environment.

I'd say if the tests that were there before did exactly what this one is (just check if instantiating does not break, and nothing else) then we could probably remove them? Obviously not if they also had more advanced checks.

xtrack/line.py
```diff
@@ -65,10 +65,11 @@
     xt.UniformSolenoid, xt.Solenoid, xt.VariableSolenoid,
     xt.Cavity, xt.RFMultipole, xt.CrabCavity, xt.ReferenceEnergyIncrease,
     xt.ReferenceEnergyChange,
-    xt.Translation, xt.Rotation, xt.XRotation, xt.TimeDelay,
```
szymonlopaciuk, line 68: Nice find (:

szymonlopaciuk (approved):
Looks good to me

JPTS2:
Updated now to answer the concerns of @giadarol!
Some interesting things surfaced (and were already missing for the Elens and SimpleThinQuadrupole if using env.set
Now fixed

The env.new parsing step now checks if an xo.Field is an array under the hood
It also no longer assumes 1D arrays if so, but will walk through the full set

Tests are done with SecondOrderTaylorMap because it's the only element I can see that has 1, 2 and 3D arrays
This doesn't affect anything like knl and ksl, because those are zero by default it's easier
This only matters when we want to initialise arrays with non zero values (like the I matrix for the Taylor maps)

xtrack/environment.py
```diff
@@ -2477,34 +2487,88 @@ def apply_filling_pattern(self, filling_pattern_cw, filling_pattern_acw,
 )
 
 
+def _parse_array_value(vv, _eval, index=()):
```
szymonlopaciuk: I'd rename `_eval` to `eval_`, since it's not really private. On the other hand `index` could be `_index`?

xtrack/environment.py
```diff
@@ -2312,11 +2312,21 @@ def _set_kwargs(self, name, ref_kwargs, value_kwargs, container, container_refs,
...
                 if kk in ref_kwargs or not isinit:
-                    for ii, vvv in enumerate(value_kwargs[kk]):
-                        if ref_kwargs[kk][ii] is not None:
-                            getattr(container_refs[name], kk)[ii] = ref_kwargs[kk][ii]
+                    attr_ref = getattr(container_refs[name], kk)
+                    leaves = ref_kwargs[kk]
+                    if not isinstance(leaves, list):
+                        # Whole array passed as one reference: wire item by item.
+                        leaves = [((ii,), leaves[ii]) for ii in range(len_value)]
+                    for index_tuple, ref in leaves:
```
szymonlopaciuk: I'm probably missing something, but I'm struggling to understand.

Say I have `knl = [1, 2, '2 * x', 4]`, I guess `leaves` for `kk = 'knl'` will be `[1, 2, '2 * x', 4]`, right? Then this iteration does not make sense, unless there's a hidden contract for `ref_kwargs` that is not obvious here?

xtrack/environment.py
```diff
@@ -2477,34 +2487,88 @@ def apply_filling_pattern(self, filling_pattern_cw, filling_pattern_acw,
...
+        return vv._value, [(index, vv)]
+    if isinstance(vv, str):
+        ref = _eval(vv)
+        value = ref._value if hasattr(ref, '_value') else ref
+        return value, [(index, ref)]
+    if hasattr(vv, '__iter__') and not isinstance(vv, (str, bytes)):
+        values = []
+        leaves = []
+        for ii, item in enumerate(vv):
+            sub_value, sub_leaves = _parse_array_value(item, _eval, index + (ii,))
```
szymonlopaciuk: On one hand you could avoid the recursion by using `np.ndindex`; very similar thing here:

https://github.com/xsuite/xtrack/blob/35f2c413ac77a63c550ff94a608a0ce4c8a9219f/xtrack/mad_loader.py#L63-L75

On the other hand, this allows us to parse lists of lists that have non-homogeneous dimensions.

I wonder if the latter is ever needed.

xtrack/environment.py
```diff
@@ -2477,34 +2487,88 @@ def apply_filling_pattern(self, filling_pattern_cw, filling_pattern_acw,
...
+            values.append(sub_value)
+            leaves.extend(sub_leaves)
+        return values, leaves
+    return vv, [(index, None)]
+
+
 def _parse_kwargs(cls, kwargs, _eval):
```
szymonlopaciuk: Oh, now I see that `_eval` wasn't you.

Still, if you can, ditto about `_eval`.

xtrack/environment.py
```diff
@@ -2477,34 +2487,88 @@ def apply_filling_pattern(self, filling_pattern_cw, filling_pattern_acw,
...
     value_kwargs = {}
     for kk in kwargs:
         if hasattr(kwargs[kk], '_value'):
             ref_kwargs[kk] = kwargs[kk]
             value_kwargs[kk] = kwargs[kk]._value
         elif (hasattr(cls, '_xofields') and kk in cls._xofields
-                and xo.array.is_array(cls._xofields[kk])):
+                and xo.array.is_array(
+                    cls._xofields[kk].ftype
+                    if isinstance(cls._xofields[kk], xo.Field)
```
szymonlopaciuk: Not the easiest read, especially in the context of the whole function, though mostly it's not your doing.

I thought perhaps too much about how to improve this, but did not come up with any simple solutions.

Maybe just add a comment "kk defines an array type field on the hybrid class cls" or something to that effect.

szymonlopaciuk (changes requested):

xtrack/line.py
JPTS2: Thank you!

tests/test_environment.py
JPTS2: It was less that some were rigorously tested and others excluded, but test_env_new does some testing of things like replicas etc and ends up covering bends, some apertures, markers etc

xtrack/environment.py
JPTS2: `eval_` done. On `index`: that argument no longer exists — switching to `np.ndindex` means the indices are generated rather than threaded through, so there is nothing to hide.
What is left is an ordinary loop variable, where a leading underscore would suggest "don't pass this" about something that cannot be passed.

xtrack/environment.py
JPTS2: Done there too.
Checked it isn't public first: neither function is used outside xtrack, and `_eval` is never passed as a keyword, so positional callers are unaffected. Matched on word boundaries so `_xdeps_eval` was left alone.

xtrack/environment.py
JPTS2: You weren't missing anything: it was implicit not explicit.
For `knl = [1, 2, '2 * x', 4]` it wasn't the values — it was one entry per item: `[((0,), None), ((1,), None), ((2,), <ref>), ((3,), None)]`.

The problem was that `ref_kwargs` held two different shapes: a single reference for scalar fields and for a whole array passed as one reference, or that list for array fields — so `_set_kwargs` had to check which it had.
Array fields now always produce the list, and that branch is gone.
Entries also carry their own value now, `(index, ref_or_None, item)`, so the consumer no longer re-walks the array to recover something already known.

xtrack/environment.py
JPTS2: Switched to `np.ndindex`, thanks — it also hands back the index tuples directly, which is exactly what's needed here.

On whether ragged is ever needed: no. xobjects array fields are rectangular by construction, including dynamic-shape ones like `xo.Float64[:, :, :, :]` where the dimensions are runtime-determined but the array is still regular, so a ragged nested list could never be assigned to one of these fields.
Non-square shapes like `[4, 3]` were never the issue either way.
It is now rejected explicitly with an error naming the field, rather than silently mis-parsed.

This walk substantially duplicates `get_value` in `mad_loader.py`, and none of it is really xtrack-specific. It may belong in xdeps?

xtrack/environment.py
JPTS2: Reworked rather than commented: the field type is resolved once into a local at the top of the loop, so the condition is now a plain `is_array(field_type)`.
Also took `cls`/`kk` to `hybrid_class`/`field_name` in both functions while in there.

JPTS2:
Pushed addressing all of the above — replies on the individual threads.

From re-reading my own diff: I'd dropped the "should be an iterable" error while restructuring now restored and
tested.
The knl/ksl length guard also had an operator-precedence bug so it never applied to knl (bracketed — unreachable in practice.

These issues were all caught by manual review rather than the suite, which stayed green throughout...
A second pair of eyes is therefore useful!

xtrack/environment.py
szymonlopaciuk: I was thinking the same actually... I think there's a few helpers we could offload, `get_value` (used countless times everywhere in xtrack), but here also `is_hybrid_class` would be helpful instead of the repeated `hasattr(typ, '_xofields')`... Maybe we can open and issue and note down where this could be useful

szymonlopaciuk (approved):
Looks good, thanks for the changes!

PR #838, opened 2026-08-10
Refactor/twiss

giadarol:


examples/fcc_ee_solenoid/009_parametric_version.py
```diff
@@ -1,6 +1,8 @@
 # Based on E. Boscolo, A. Ciarma, E. Burkhardt, https://cds.cern.ch/record/2948247
 # Nuclear Instruments and Methods in Physics Research A 1083 (2026) 171135
 
+from tkinter import TRUE
```
szymonlopaciuk, line 4: I think this might be unused

examples/twiss/refactor.md
szymonlopaciuk, line 1: This probably should be removed from the release.

xtrack/twiss/__init__.py
```diff
@@ -0,0 +1,82 @@
...
+    compute_R_matrix,
+    compute_T_matrix_line,
+    find_closed_orbit_line,
+    get_R_matrix,
+    get_T_matrix_line,
+    get_non_linear_chromaticity,
+    twiss_line,
+):
+    _public_obj.__module__ = __name__
+del _public_obj
```
szymonlopaciuk, line 54: Maybe could be derived from `__all__` so that it doesn't get forgotten about?

xtrack/twiss/twiss_defaults_and_input_preparation.py
```diff
@@ -0,0 +1,559 @@
...
+        num_turns,
+        disable_apertures,
+):
+    return (
+        (step_W_sigma or 0.01),
+        (nemitt_x or 1e-6),
+        (nemitt_y or 1e-6),
+        (delta_disp or 1e-5),
+        (delta_chrom or 5e-5),
+        (zeta_disp or 1e-3),
```
szymonlopaciuk, line 74: Need to be careful here that `0.0` is not a reasonable user input: I think for most it's not but I'm not sure about all of them.

xtrack/twiss/twiss_defaults_and_input_preparation.py
```diff
@@ -0,0 +1,559 @@
...
+    'dx', 'dpx', 'dy', 'dpy', 'dzeta',
+    'mux', 'muy', 'muzeta',
+    'ax_chrom', 'bx_chrom', 'ay_chrom', 'by_chrom',
+    'ddx', 'ddpx', 'ddy', 'ddpy',
+]
+
+
+def _apply_twiss_defaults(
+        *,
+        step_W_sigma,
```
szymonlopaciuk, line 38: This method 'effectively' takes a dict (through `**kwargs` so it admittedly does indirectly check structure) but outputs a tuple. This makes it kind of verbose for what it does and the call site is messy as well...

I read in the design doc there was an idea for a `TwissConfig` dataclass-type object. I think passing something like that here wouldn't be a bad idea as validity of input names would be checked and we wouldn't need to deconstruct the tuple into a dict later on.

xtrack/twiss/twiss_defaults_and_input_preparation.py
```diff
@@ -0,0 +1,559 @@
...
+            if compute_lattice_functions is not None else True),
+        (chrom if chrom is not None else None),
+        (num_turns or 1),
+        (disable_apertures if disable_apertures is not None else True),
+    )
+
+
+def _normalize_twiss_inputs(twiss_kwargs):
+
+    import xtrack as xt  # Local import avoids circular imports.
```
szymonlopaciuk, line 104: Are `xt.START` and `xt.END` used in other contexts than twiss? If not then they could become the responsibility of this module, and this import could go.

xtrack/twiss/twiss_defaults_and_input_preparation.py
```diff
@@ -0,0 +1,559 @@
...
+        assert twiss_config['init'] in ('periodic', 'full_periodic')
+    if (not twiss_config['periodic']
+            and (twiss_config['delta0'] is not None or twiss_config['zeta0'] is not None)):
+        raise ValueError(
+            'delta0 and zeta0 cannot be provided for open twiss')
+
+
+def _resolve_twiss_range_endpoint(line, endpoint, reverse):
+
+    import xtrack as xt  # Local import avoids circular imports.
```
szymonlopaciuk, line 326: Ditto

xtrack/twiss/twiss_defaults_and_input_preparation.py
szymonlopaciuk, line 1: Overarching comment on file: I think having a dataclass collecting twiss parameters instead of a dictionary, and passing that around, could help avoid sneaking in any typos later on, would help static checkers, and probably make it look a bit nicer (less dict-key accesses).

In fact a lot of these functions are simply methods of such a construct.

xtrack/twiss/twiss_defaults_and_input_preparation.py
```diff
@@ -0,0 +1,559 @@
...
+        only_orbit=twiss_kwargs['only_orbit'],
+        compute_R_element_by_element=twiss_kwargs['compute_R_element_by_element'],
+        compute_lattice_functions=twiss_kwargs['compute_lattice_functions'],
+        chrom=twiss_kwargs['chrom'],
+        num_turns=twiss_kwargs['num_turns'],
+        disable_apertures=twiss_kwargs['disable_apertures'],
+    )
+
+    if twiss_kwargs['only_markers']:
+        raise NotImplementedError('``only_markers`` not supported anymore')
```
szymonlopaciuk, line 252: Could go first? Then `_apply_twiss_defaults` won't need to worry about it.

xtrack/twiss/twiss_defaults_and_input_preparation.py
```diff
@@ -0,0 +1,559 @@
...
+
+def _prepare_twiss_at_s_markers(twiss_config):
+
+    if twiss_config['at_s'] is None:
+        return
+
+    if twiss_config['reverse']:
+        raise NotImplementedError('``at_s`` not implemented for ``reverse``=True')
+    if np.isscalar(twiss_config['at_s']):
+        twiss_config['at_s'] = [twiss_config['at_s']]
```
szymonlopaciuk, line 362: If `at_s` can be a numpy array anyway, `np.atleast_1d` does the same job

xtrack/twiss/twiss_defaults_and_input_preparation.py
```diff
@@ -0,0 +1,559 @@
...
+    auxline.insert(insertions)
+
+    auxtracker = xt.Tracker(
+        _buffer=tracker._buffer,
+        io_buffer=tracker.io_buffer,
+        line=auxline,
+        particles_monitor_class=None,
+    )
+    auxtracker.line.config = tracker.line.config.copy()
+    auxtracker.line._extra_config = tracker.line._extra_config.copy()
```
szymonlopaciuk, line 410: This is quite manual but looks similar to

```python
auxline = tracker.line.copy()
... # Insert the markers
auxline.build_tracker()
return auxline.tracker, names_inserted_markers
```

Maybe it could be helpful to explain why we go the manual route here

xtrack/twiss/twiss_init.py
```diff
@@ -0,0 +1,699 @@
...
+            #     dpx=self._temp_optics_data['dpx'],
+            #     dpy=self._temp_optics_data['dpy'] * (-1 if input_reversed else 1),
+            #     )
+            # aux_line = xt.Line(elements=[aux_segment])
+            # aux_line.particle_ref = particle_on_co.copy(
+            #                             _context=xo.context_default)
+            # aux_line.particle_ref.reorganize()
+            # aux_line.build_tracker()
+            # aux_tw = aux_line.twiss()
+            # W_matrix = aux_tw.W_matrix[0]
```
szymonlopaciuk, line 259: Do we need to keep it?

xtrack/twiss/twiss_init.py
```diff
@@ -0,0 +1,699 @@
...
+        if name in self.__dict__:
+            return self.__dict__[name]
+        elif hasattr(self.__dict__['particle_on_co'], name):
+            # e.g. tw_init['x'] returns tw_init.particle_on_co.x
+            out = getattr(self.__dict__['particle_on_co'], name)
+            #always cpu
+            if hasattr(out, 'get'):
+                out = out.get()
+            if hasattr(out, '__iter__'):
+                out = out [0]
```
szymonlopaciuk, line 373: Stray space

xtrack/twiss/twiss_init.py
```diff
@@ -0,0 +1,699 @@
...
+    'step_W_sigma',
+    'compute_R_element_by_element',
+    'only_markers',
+    'only_orbit',
+    'periodic_mode',
+    'include_collective',
+)
+
+
+def _build_twiss_init_from_inputs(twiss_config):
```
szymonlopaciuk, line 589: I think this would make a nice factory method on `TwissInit`

xtrack/twiss/twiss_postprocess_and_complem_results.py
```diff
@@ -0,0 +1,358 @@
...
+
+    twiss_res._data['R_matrix'] = twiss_config['R_matrix']
+    twiss_res._data['steps_R_matrix'] = twiss_config['steps_R_matrix']
+    twiss_res._data['steps_r_matrix'] = twiss_config['steps_R_matrix']  # deprecated
+    twiss_res._data['R_matrix_ebe'] = twiss_config['RR_ebe']
+
+    line = twiss_config['line']
+    method = twiss_config['method']
+    s_vect = twiss_res['s']
+    line_length = line.tracker._tracker_data_base.line_length
```
szymonlopaciuk, line 34: This seems like a roundabout way of obtaining this quantity. Maybe `Line.get_length()` could read this in the case when the tracker is built?

xtrack/twiss/twiss_postprocess_and_complem_results.py
```diff
@@ -0,0 +1,358 @@
...
+    part_on_co = twiss_res['particle_on_co']
+    W_matrix = twiss_res['W_matrix']
+
+    beta0 = part_on_co._xobject.beta0[0]
+    gamma0 = part_on_co._xobject.gamma0[0]
```
szymonlopaciuk, line 39: Why not directly `part_on_co.beta0[0]` (same for `gamma0`)?

xtrack/twiss/twiss_postprocess_and_complem_results.py
```diff
@@ -0,0 +1,358 @@
...
+        num_turns=twiss_config['num_turns'],
+        hide_thin_groups=twiss_config['hide_thin_groups'],
+        only_markers=twiss_config['only_markers'],
+        periodic=twiss_config['periodic'],
+        periodic_mode=twiss_config['periodic_mode'],
+        include_collective=twiss_config['include_collective'],
+    )
+    twiss_res._data.update(cols_chrom)
+    twiss_res._data.update(scalars_chrom)
+    twiss_res._col_names += list(cols_chrom.keys())
```
szymonlopaciuk, line 190: I'm dreaming of something like:

```python
table_chrom = twiss_config.get_chromatic_functions()
twiss_res = twiss_res | table_chrom
# Or something like Table.update, that way we wouldn't need to
# rely on the internal implementation of Table. I think I've done 
# something like this before, so maybe extending xd.Table to 
# support this would be useful. Maybe not right now, but something
# to think about.
```

xtrack/footprint.py
```diff
@@ -2,6 +2,13 @@
 import xobjects as xo
 import xtrack as xt
 
+
+def _trapezoid(nplike_lib, y, x=None, axis=-1):
+    if hasattr(nplike_lib, 'trapezoid'):  # numpy >= 2.0
+        return nplike_lib.trapezoid(y, x=x, axis=axis)
+    return nplike_lib.trapz(y, x=x, axis=axis)
```
szymonlopaciuk, line 9: Very similar thing in `chromatic_functions.py`. Maybe an opportunity to unify in a single place (like `general`)?

xtrack/twiss/twiss.py
```diff
@@ -0,0 +1,637 @@
...
+            twiss_config['init'] = _build_twiss_init_from_inputs(twiss_config)
+            result_init = twiss_config['init'].copy()
+            _clear_twiss_init_input_fields(twiss_config)
+
+            crosses_line_boundary, init_is_at_boundary = (
+                _get_open_twiss_range_flags(twiss_config))
+            if not crosses_line_boundary and init_is_at_boundary:
+                twiss_res = _compute_base_twiss(twiss_config)
+            else:
+                twiss_res = _handle_init_inside_range_and_line_wrap(
```
szymonlopaciuk, line 403: `_twiss_with_init_inside_range_and_line_wrap`? My first instinct was 'where is the twiss calculated?' and only then I saw the function takes the twiss calculator.

xtrack/twiss/twiss.py
```diff
@@ -0,0 +1,637 @@
...
+            track_flag_updates=track_flag_updates,
+            line_config_updates=line_config_updates,
+            freeze_longitudinal=twiss_config['freeze_longitudinal'],
+            freeze_energy=twiss_config['freeze_energy'])
+
+        # Determine the route (periodic / open / periodic_one_turn_custom_start
+        # / open_one_turn_custom_start / open_init_from_full_periodic)
+        route = _select_twiss_route(twiss_config)
+
+        if route == 'periodic':
```
szymonlopaciuk, line 385: We're now on Python 3.11+ which means we can finally use match-case[^1], if you think it's appropriate anyway.

[^1]: https://docs.python.org/3/tutorial/controlflow.html#match-statements

xtrack/twiss/twiss.py
```diff
@@ -0,0 +1,637 @@
...
+            assert start_index >= end_index, (
+                'start must be at or after end in reverse mode')
+        else:
+            assert start_index <= end_index, (
+                'start must be at or before end in forward mode')
+
+    if reverse:
+        start, end = end, start
+    twiss_config['start'] = start
+    twiss_config['end'] = end
```
szymonlopaciuk, line 528: This could be inside the if, right? The values of start and end don't seem to change anywhere else before.

xtrack/twiss/chromatic_functions.py
```diff
@@ -0,0 +1,252 @@
...
+        # y = on_momentum_twiss_res.y
+        # py = on_momentum_twiss_res.py
+        # ddqx = (tw_chrom_res[1].mux[-1] - 2 * mux[-1] + tw_chrom_res[0].mux[-1]
+        #         ) / delta_chrom**2
+        # ddqy = (tw_chrom_res[1].muy[-1] - 2 * muy[-1] + tw_chrom_res[0].muy[-1]
+        #         ) / delta_chrom**2
+        # ddx = (tw_chrom_res[1].x - 2 * x + tw_chrom_res[0].x) / delta_chrom**2
+        # ddpx = (tw_chrom_res[1].px - 2 * px + tw_chrom_res[0].px) / delta_chrom**2
+        # ddy = (tw_chrom_res[1].y - 2 * y + tw_chrom_res[0].y) / delta_chrom**2
+        # ddpy = (tw_chrom_res[1].py - 2 * py + tw_chrom_res[0].py) / delta_chrom**2
```
szymonlopaciuk, line 246: Do we need it, and if so maybe add a comment explaining why (not yet tested?)

xtrack/twiss/chromatic_functions.py
```diff
@@ -0,0 +1,252 @@
...
+    if only_markers:
+        raise NotImplementedError('only_markers not supported anymore')
+
+    if tw_chrom_res is None:
+        tw_chrom_res = []
+        for dd in [-delta_chrom, delta_chrom]:
+            tw_init_chrom = init.copy()
+
+            if periodic:
+                slip_factor_dzeta_ddelta = on_momentum_twiss_res.slip_factor_dzeta_ddelta
```
szymonlopaciuk, line 44: Quite deeply nested, maybe the whole thing, or the branches could be extracted?

xtrack/twiss/handle_init_inside_range_and_line_wrap.py
```diff
@@ -0,0 +1,261 @@
...
+    line = kwargs['line']
+    reverse = kwargs['reverse']
+
+    # Confirm that the requested traversal crosses the physical line boundary.
+    if one_turn_from_start:
+        assert start == end
+    elif not reverse:
+        assert (_element_ref_to_index(line, end)
+                < _element_ref_to_index(line, start)), (
+            'This function should not have been called')
```
szymonlopaciuk, line 76: Unlikely, but if the user ever gets this it might be confusing. I'd indicate "internal error" and to contact us?

xtrack/twiss/lattice_functions_from_W.py
```diff
@@ -0,0 +1,245 @@
...
+    sign_x = np.sign(betx)
+    sign_y = np.sign(bety)
+    betx *= sign_x
+    alfx *= sign_x
+    gamx *= sign_x
+    bety *= sign_y
+    alfy *= sign_y
+    gamy *= sign_y
+
+    return betx, alfx, gamx, bety, alfy, gamy, bety1, betx2, alfy1, alfx2, gamy1, gamx2
```
szymonlopaciuk, line 245: In places like this we could return a `NamedTuple`, makes it lighter on the eyes and less error prone at the call site.

xtrack/twiss/non_linear_chromaticity.py
```diff
@@ -0,0 +1,61 @@
...
+    out_data['dnqx'] = dnqx
+    out_data['dnqy'] = dnqy
+
+    out_data['delta0'] = delta0
+    out_data['qx'] = qx
+    out_data['qy'] = qy
+    out_data['momentum_compaction_factor'] = momentum_compaction_factor
+    out_data['twiss'] = twiss
+
+    out = xt.Table(data = out_data, index='delta0',
```
szymonlopaciuk, line 58: `data=out_data`

xtrack/twiss/spin.py
```diff
@@ -0,0 +1,346 @@
...
+    pp.spin_y = sy
+
+    line.track(pp)
+
+    return np.array([pp.spin_x[0] - sx,
+                        pp.spin_y[0] - sy,
+                        pp.spin_z[0] - sz])
+
+
+def _get_spin_polarization(tw, line, method):
```
szymonlopaciuk, line 58: The name implies there is a return value of some sort, but in reality this edits `tw` in place.

szymonlopaciuk (commented):
Looks very good! My comments are mostly surface because I don't have a complete understanding in my head of how all the internals work together. Although, after looking though this, it's a much better picture than before!

Something I keep wondering about is this: the module is mostly structured as a collection of methods, and there is a lot of passing of the state between them (dicts, tuples, lines, etc.). I made a comment about making `TwissConfig`, but I wonder if there are other opportunities to identify other collections of methods operating on the same state, that could be similarly extracted into an object (there is some post-processing of a `twiss_res` at one point for example).

I know it's hard, and maybe not relevant even, because at the end of the day if we had a `Twiss` object, then we would go back to having a 6k long file. But maybe there is a middle ground to try to reduce the passing of dicts or long lists of parameters in the internal part?

xtrack/twiss/twiss_defaults_and_input_preparation.py
giadarol: START and END are used also in the match module

examples/fcc_ee_solenoid/009_parametric_version.py
giadarol: removed

examples/twiss/refactor.md
giadarol: removed

xtrack/twiss/__init__.py
giadarol: I removed the whole loop

xtrack/twiss/twiss_defaults_and_input_preparation.py
giadarol: Valid concern, luckily they are all fine :)

xtrack/twiss/twiss_defaults_and_input_preparation.py
giadarol: change the implementation to alter the dictionary in place. The call became a one liner.

xtrack/twiss/twiss_defaults_and_input_preparation.py
giadarol: True, at the moment twiss_config is a non structured bag of data. We could further structure it in the future....

xtrack/twiss/twiss_defaults_and_input_preparation.py
giadarol: ditto

xtrack/twiss/twiss_defaults_and_input_preparation.py
giadarol: yes, moved above

xtrack/twiss/twiss_defaults_and_input_preparation.py
giadarol: `at_s` is deprecated, I don't touch it. looking forward to the day we will remove it.

xtrack/twiss/twiss_defaults_and_input_preparation.py
giadarol: `at_s` is deprecated, I don't touch it. looking forward to the day we will remove it.

xtrack/twiss/twiss_init.py
giadarol: removed

xtrack/twiss/twiss_init.py
giadarol: removed

xtrack/twiss/twiss_init.py
giadarol: as twiss_config is not part of the API, I prefer to keep this function private

xtrack/twiss/twiss_postprocess_and_complem_results.py
giadarol: on OpenCL you would need to do `.get()`

xtrack/twiss/twiss_postprocess_and_complem_results.py
giadarol: I would prefer `_get_chromatic_functions(twiss_config, twiss_res)` as it uses also data in `twiss_res` not only in `twiss_config`.

`twiss_res = twiss_res | table_chrom` would indeed be cool

xtrack/footprint.py
giadarol: or even in xobjects....

xtrack/twiss/twiss.py
giadarol: renamed to `_compute_twiss_handling_init_inside_range_and_line_wrap` long, but explicit

xtrack/twiss/spin.py
giadarol: renamed to `_add_spin_polarization`

xtrack/twiss/chromatic_functions.py
giadarol: removed

xtrack/twiss/non_linear_chromaticity.py
giadarol: Clean up

xtrack/twiss/lattice_functions_from_W.py
giadarol: I agree, although changing it in one place only would look strange, so I won't do it now. We should in general review the data flow management across the module, as mentioned in previous comments.

xtrack/twiss/handle_init_inside_range_and_line_wrap.py
giadarol: I went for: 'Corrupted input, internal error'

xtrack/twiss/chromatic_functions.py
giadarol: True. I extracted the content of the content of the long if/else body into two separate functions. It looks more readable now.

xtrack/footprint.py
giadarol: actually numpy 2.0 is out since more than two years. And numpy 1.X is not anymore supported. We should probably enforce numpy >= 2.0 and get rid of this logic

giadarol: I opened a separate issue: https://github.com/xsuite/xsuite/issues/903

xtrack/twiss/twiss.py
giadarol: I don't think it adds much in this case though...

xtrack/twiss/twiss.py
giadarol: yes, fixed

PR #839, opened 2026-08-10
Improve bend stability + backtracking through full fringe

jgray-19:
# Description

Two related fixes to thick bends: backtracking through the non-linear ("full") dipole edge, and loss of precision in weak/small-angle bends.

## Backtracking through the full dipole edge.

DipoleFringe_track_single_particle now takes a backtrack flag and shares its coefficient evaluation (DipoleFringe_coefficients) with the forward map. 

The wedge, quadrupole wedge and y-rotation are inverted by flipping the sign of the face angle, so the edge tracker folds the direction into a single sign and keeps one copy of the map ordering per face rather than one per (face, direction). 

## Backtracking of the multipole fringe

A similar fix has been implemented for the ability to backtrack through the multipole fringe. 

See https://github.com/xsuite/xsuite/pull/920#issue-5367862553 for detailed mathematical description of the fringe changes.

## Numerical stability of weak bends.

- The straight exact bend advanced `x` as `(new_pz - pz)/k0` — a difference of two square roots close to `1 + delta`, divided by a small number. Multiplying through by the conjugate gives `s*(new_px + px)/(new_pz + pz)`, in which k0 cancels exactly and the error stops growing as the bend gets weaker. 
- The y advance and path lengthening used `asin(A*px) - asin(A*new_px)`, which is replaced by a single `atan2` of the same angle difference.
- The RBend straight-body offsets `_x0_in`/`_x0_mid`/`_x0_out` (both the Python properties and the C in track_magnet) evaluated `(1 - cos(angle/2))/h` and `(sqrt_mid - cos_theta)/h`, both cancellations amplified by the large `1/h`. Using `1 - cos(u) = 2 sin(u/2)^2` and rationalising the square-root differences makes h cancel analytically.

This image shows what my stability changes do for the Rbends - could impact FCC. 
006_noise_reduction_changes

## Tests

| Test | File | Status |
| --- | --- | --- |
| `test_backtrack_with_bend` | `tests/test_elements_thick.py` | new |
| `test_backtrack_with_quadrupole` | `tests/test_elements_thick.py` | new |
| `test_backtrack_with_cfm` | `tests/test_elements_thick.py` | new |
| `test_backtrack_with_dipole_edge` | `tests/test_elements_thick.py` | new |
| `test_rbend_full_edge_backtracks_many_coordinates` | `tests/test_magnet.py` | new |
| `test_fringe_backtrack_with_exactly_zero_coordinate` | `tests/test_magnet.py` | new |
| `test_edge_multipole_fringe_without_dipole_component` | `tests/test_magnet.py` | modified — gained a backtrack round trip |
| `test_drift_kick_drift_exact_converges_to_straight_body_rbend` | `tests/test_rbend_straight_body.py` | new |
| `test_multipole_edge_backtracks_many_coordinates` | `tests/test_magnet.py` | new |
| `test_magnet_edge_multipole_backtracks_with_integrated_strengths` | `tests/test_magnet.py` | new |
| `test_multipole_fringe_backtrack_not_converged` | `tests/test_magnet.py` | new |

Checklist

Mandatory:

Optional:

- [x] The code I wrote follows good style practices (see PEP 8 and PEP 20).
- [x] I have updated the docs in relation to my changes  (see https://github.com/xsuite/xsuite/pull/920)
- [ ] I have tested also GPU contexts - I can't here.


giadarol:
Thank you!!!
Could you please open a sister PR in xsuite updating the formulas in the Physics manual with these manipulations?

rdemaria:
If I recall well the fringe has also a cubic term. I cannot check easily now, have you changed the forward map of the fringe?

jgray-19:
> If I recall well the fringe has also a cubic term. I cannot check easily now, have you changed the forward map of the fringe?

I did not change the forward map of the fringe. I refactored it in the latest commit, but in the commits before, you can see there is no modification of the forward part.

jgray-19:
I have updated the description to be more detailed and precise by what I meant by the quadratic. Some form of this will go into the physics manual.

giadarol:
Looks very neat.

I was thinking that, as performance is not very critical in backtracking mode, we could actually afford a proper iteration with a convergence check to exit the loop (anyhow it seems to converge very quickly). And, if convergence is not reached within a reasonable number of iterations (say MAX_ITER=10), we could kill the particle with an error code.

Like this if some user comes up with parameters/coordinates for which we do not converge, we would detect it and not continue silently.

What do you think?

rdemaria:
The check should not be expensive because it is only one scalar equation.

In #810 there a non-iterativive version that should give a good compromise between accuracy and speed. We can discuss next week, if you are interested.

PR #844, opened 2026-08-14
Utilise inserting a list of xt.places for synctime installation

JoshuaAppleby:
## Description
When creating SyncTime elements and adding them to the line, it currently results in a long stream of slicing line readouts. It is cleaner to create all SyncTime elements, create the place objects and then insert them all in one go.

Note: This may be improved later by using env.new instead of env["element_name"] = ... when SyncTime is in _ALLOWED_ELEMENT_TYPES_IN_NEW in xt.environment.py

Unchanged functionality on user side.


giadarol:
great! thank you!

giadarol (approved):

PR #848, opened 2026-08-17
Large improvement in time needed for particles creation

freddieknets:
## Description

Particles generation can take a long time.
For instance, generating 100M particles on my machines (same time on a MacBook M3 and AMD Ryzen 7 9850X3D) takes a bit more than 30 seconds.
This PR addresses this in two steps, and in doing so, gains more than an order of magnitude in speed (from 30 to 2.2 seconds).

### PDG optimisation

Numeric PDG-ID arrays were processed recursively in Python—one function call and several type checks per particle.
Ion lookup scanned and lowercased all 236 short/long element names. Initial mass-table construction triggered this hundreds of times.
The new approach:
 - Converts numeric arrays through NumPy; int64 arrays are zero-copy
 - Parses the ion’s mass number, then uses a precomputed lowercase element dictionary for O(1) lookup
 - Retains existing string, antiparticle, and ion formats

Very strong speed gain: from 30 seconds down to 10 seconds.

### Particles `__init__` optimisation

Key improvements:
 - Avoid sentinel-prefilling arrays when every slot will be overwritten
 - Avoid boolean masks and masked copies for all-valid particles
 - Broadcast CPU scalar inputs without allocating full-size temporary arrays
 - Avoid unnecessary NumPy copies and eager `particle_id` allocation
 - Skip consistency checks when input and computed value are identical
 - Replace Python-level mass-ratio iteration with vectorised checks
 - Skip unnecessary default-zeta calculations and redundant CPU reorganisation
 - Fresh CPU/CUDA buffers are already zeroed, so initialisation skips redundant zero writes for default fields, `pdg_id`, and random state; reused buffers, pyopencl, extra capacity, invalid particles, and signed -0.0 retain the original handling
 - `_update_energy_deviations` avoids unused `pzeta` calculations and reuses the already-computed `ptau` expression for `rvv`, eliminating a second square root on the physical CPU path; unusual/negative-energy inputs use the original formula
 - `_update_zeta` no longer derives `tau` unless supplied and returns immediately for defaults

Further speed gain: from 10 seconds down to 2.2 seconds.

All Xtrack tests pass on CPU; currently running CUDA tests overnight.

## Checklist

Mandatory: 

Optional:

### DIFF
--- a/xtrack/particles/particles.py
+++ b/xtrack/particles/particles.py
@@ -2055,26 +2055,13 @@ class Particles(xo.HybridClass):
                                     mask=mask)
 
         delta = self._delta  # Cupy complains if we later assign LinkedArray
-        can_reuse_ptau = False
-        # A device-side reduction would force synchronization, so retain the
-        # original expression on GPU contexts.
-        if isinstance(self._context, xo.ContextCpu):
-            one_plus_ptau_beta0 = 1 + _ptau * beta0
-            values_to_check = (one_plus_ptau_beta0 if mask is None
-                               else one_plus_ptau_beta0[mask])
-            can_reuse_ptau = bool(
-                self._context.nplike_lib.all(values_to_check >= 0))
-        if can_reuse_ptau:
-            # This is algebraically identical to recovering ptau * beta0
-            # from delta through another square root, in the physical
-            # positive-energy branch. Reusing ptau avoids that round trip.
-            rvv_denominator = one_plus_ptau_beta0
-        else:
-            delta_beta0 = delta * beta0
-            ptau_beta0 = _sqrt(
-                delta_beta0 ** 2 + 2 * delta_beta0 * beta0 + 1) - 1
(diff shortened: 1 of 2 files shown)


xtrack/particles/particles.py
```diff
@@ -666,20 +690,24 @@ def __init__(
...
-                kwargs[field] = np.array(kwargs[field])
+                kwargs[field] = np.asarray(
+                    kwargs[field], dtype=xotype._dtype)
 
-            # Coerce the right type so that we can allocate the right array
-            # in the target context. PyOpenCL gets fussy if types don't match
-            # in calculations.
-            if kwargs[field].dtype != xotype._dtype:
-                kwargs[field] = kwargs[field].astype(xotype._dtype)
+            # PyOpenCL gets fussy if types don't match in calculations.
```
giadarol, line 710: we skip the type casting here?

PR #859, opened 2026-09-17
Update synctime to have standard beam element attributes

### LINKED ISSUE #913: SyncTime has no attribute "to_dict"
When copying an environment that has SyncTime elements: AttributeError: 'SyncTime' object has no attribute 'to_dict'

Using: env.copy()

JoshuaAppleby:
Fixes xsuite/xsuite#913

So env.copy() can happen with synctime elements included.


JoshuaAppleby:
My only concern is the attributes.

iscollective = True
allow_track = False

Please confirm/change before merge!

